PR #21, opened 2021-08-06
Set some code standards and other improvements.

drgrice1:
Add vuex-persistedstate to save the vuex session state to session
storage.  This makes page refresh work (and has other possibile uses).

Remove the webwork3 url prefix from the code, and instead put that into
the publicPath option in quasar.conf.js.

Add some eslint rule for syntax.  These are rules that I prefer, but are
up for discussion.  One thing that this does is disables console.log's
for the production build.  Note that these should also not be committed
to git (they can be committed commented out if the must be there).  All
existing console.log's have been removed from the code.

Add the .editorconfig file.  This is honored by many text editors by
default.  Some require a plugin (for example vim).  See
https://editorconfig.org/ for details.

Add stylelint to the build system.  It only emits errors for the
production build, and not for the development build.

Clean up the build system.  Mainly remove most of the comments added by
the quasar cli.  All of those are well documented on the quasar website.
Also remove things from the build system that we will not be using like
ssr and pwa.  These can be added back if they are used.

Remove the .quasar directory from git, and add it to the ignore file.
That directory is generated by the build, and its contents are different
for the development build and the production build.  It should not be in
the repository.

Switch from using the user's email address to using the user's login id
(a username).  I am not sure why the email address was used.  We need to
discuss this.

Add a method for deployment to the README.md file.

There are a few other minor improvements to the ui, but most of the
other changes are clean up and fixes for the lintian rules.


pstaabp:
@drgrice1 Overall, I think this is great.  Some of this, I haven't thought about implementing yet and I'll dive into the linting rules, but generally I'm flexible with that.  

> Switch from using the user's email address to using the user's login id
> (a username). I am not sure why the email address was used. We need to
> discuss this.

I've been wondering about this and needs more discussion.  Tani had suggested that the users table only stored the email address and use that as the login).  All other user information could be stored elsewhere. 

I've been doing some more working on the UI this week and have been basing things on the webwork2 way.

drgrice1:
> > Switch from using the user's email address to using the user's login id
> > (a username). I am not sure why the email address was used. We need to
> > discuss this.
> 
> I've been wondering about this and needs more discussion. Tani had suggested that the users table only stored the email address and use that as the login). All other user information could be stored elsewhere.

How do you create an account for a user for which you do not have an email address?  I have many students that fall into this category.  Furthermore, guest accounts will not have an email address.  I don't think there is any further discussion needed.  Using the email address for the login is not well thought out.

PR #27, opened 2021-08-19
Fix the unit tests workflow

drgrice1:
For now the code coverage part of the workflow is disabled.  I am not sure what is intended there, but it isn't working.

This uses a docker image from my repository on docker hub (drgrice1/webwork3).  We need to create a webwork repository on docker hub if we are going to use a prebuilt image in this way, rather than using one of our private repositories.


drgrice1:
Note that all workflow checks have passed for this pull request!

PR #28, opened 2021-08-19
Initial testing of i18n code

pstaabp:
This checks in the i18n code and allows dynamic language changes.  To change the language, select the new language under User Settings (dropdown from Logged in User name).


src/boot/i18n.ts
```diff
@@ -0,0 +1,15 @@
+import { boot } from 'quasar/wrappers';
+import { createI18n } from 'vue-i18n';
+import messages from 'src/i18n';
+
+const i18n = createI18n({
+	locale: 'fr',
```
drgrice1, line 6: The default here should be 'en_US'.

src/layouts/MenuBar.vue
```diff
@@ -1,43 +1,49 @@
...
+
+			<q-card-section class="q-pt-none">
+				<q-select label="Language" v-model="$i18n.locale" :options="$i18n.availableLocales" />
+			</q-card-section>
+			<q-card-actions align="right" class="bg-white text-teal">
+				<q-btn flat label="OK" v-close-popup />
+			</q-card-actions>
+		</q-card>
+	</q-dialog>
+	</div>
```
drgrice1, line 46: This completely reverts the addition of the WeBWorK and MAA logos.  It looks like you forgot to pull before starting work.  Either that or you forgot to relegate your work with the upstream changes.

src/layouts/MenuBar.vue
```diff
@@ -51,7 +57,7 @@ export default defineComponent({
 	emits: {
 		toggle: null
 	},
```
drgrice1, line 59: See there are no other events that need validation, the shorthand array form is preferable.
```suggestion
	emits: ["toggle"],
```

drgrice1 (approved):
Looks good.  But there are a few issues.  I will submit a pull request to this that adds back the logos and addresses my other comments.

src/layouts/MenuBar.vue
pstaabp: I thought I had before doing this work.  Oops

src/boot/i18n.ts
pstaabp: agreed.

PR #29, opened 2021-08-19
Fix POD issues.

drgrice1:
HTML does not work in POD.  So you can't use `<code>...</code>` tags.
Instead use `C<...>`.

Command paragraphs (`=headN`, `=item`, `=cut`, etc.) should always have an
empty line before and after them.

Command paragraphs automatically begin POD, so there is no need to have
`=pod` before `=headN`.  In fact `=pod` is just a basic command paragraph that
begins a basic unformatted section.  Furthermore, using `=pod` immediately
followed by something like `=head1` with no empty line between is an
error.

`=item`'s need to be wrapped in `=over` and `=back`.  Furthermore, items
at different levels need an addional over and back.  So `=item -`
nested inside an `=item *` needs another over and back.


drgrice1:
Ahh, I knew this would come up eventually.  Running the linter locally showed these issues.  It doesn't like the repetitive code in some of the ResultSet files.

PR #30, opened 2021-08-20
Fix POD issues.

drgrice1:
HTML does not work in POD. So you can't use `<code>...</code>` tags.
Instead use `C<...>`.

Command paragraphs (`=headN`, `=item`, `=cut`, etc.) should always have an empty line before and after them.

Command paragraphs automatically begin POD, so there is no need to have `=pod` before `=headN`. In fact `=pod` is just a basic command paragraph that begins a basic unformatted section. Furthermore, using `=pod` immediately followed by something like `=head1` with no empty line between is an error.

`=item`'s need to be wrapped in `=over` and `=back`. Furthermore, items at different levels need an additional over and back. So `=item -` nested inside an `=item *` needs another over and back.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -60,7 +60,10 @@ jobs:
       # Run Linter against code base #
       ################################
       - name: Super-Linter
-        uses: github/super-linter@v4.5.1
+        # Note that this actually uses super-linter version 4.6.0 which is the desired version.
+        # See https://github.com/github/super-linter/issues/1839.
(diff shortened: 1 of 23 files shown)


drgrice1:
The super-linter workflow now runs cleanly on the entire code base (tested locally).  So that means that failures for future pull requests are caused by that pull request.  *.pm and *.pl files are ignored by the jscpd linter though.

pstaabp:
Looks good.  Let's pull it in.

PR #31, opened 2021-08-20
A number of UI and related db improvements

pstaabp:
Improvements to mainly the Problem Set Details pages and some related UI/db.


bin/webwork3
```diff
@@ -10,5 +10,14 @@ use Mojo::File qw(curfile);
 use lib curfile->dirname->sibling('lib')->to_string;
 use Mojolicious::Commands;
 
+# check if the config file has been created
+my $webwork_dir = curfile->dirname->dirname("..");
+
+warn "The file $webwork_dir/conf/webwork3.yml does not exist.  Perhaps you haven't copied webwork3.dist.yml to this file."
```
drgrice1:
```suggestion
warn "The file $webwork_dir/conf/webwork3.yml does not exist.  Perhaps you haven't copied webwork3.yml.dist to this file."
```
The .dist extension is after the .yml extension on the actual file.  Perhaps we should switch that though so that linters and editors properly recognize the file type?

bin/webwork3
```diff
@@ -10,5 +10,12 @@ use Mojo::File qw(curfile);
 use lib curfile->dirname->sibling('lib')->to_string;
 use Mojolicious::Commands;
 
+# check if the config file has been created
+my $webwork_dir = curfile->dirname->dirname("..");
+
+warn qq!The file $webwork_dir/conf/webwork3.yml does not exist.
+	Perhaps you haven't copied webwork3.dist.yml to this file.!
```
drgrice1, line 17:
```suggestion
	Perhaps you haven't copied webwork3.yml.dist to this file.!
```
The dist file is `webwork3.yml.dist`.  Should we change the file name to `webwork3.dist.yml`?  The advantage of putting the `yml` extension at the end is that it will be recognized as a YAML file by text editors and linters and such.

package.json
```diff
@@ -52,5 +54,8 @@
 		"node": ">= 12.22.1",
 		"npm": ">= 6.13.4",
 		"yarn": ">= 1.21.1"
+	},
+	"jscpd": {
+    "threshold": 5
```
drgrice1, line 59: This needs to go in the .jscpd.json file (added in my pull request).  jscpd honors the block in the package.json file, but I don't think that super-linter does.  Furthermore, I would rather not clutter the package.json file.

src/components/admin/AddCourse.vue
```diff
@@ -86,18 +79,20 @@ export default defineComponent({
...
-			login,
+			username,
 			course_dates,
 			instructor_exists,
-			checkUser: async () => { // lookup the user by login to see if already exists
-				const _user = await store.dispatch('users/getGlobalUser', user.value.login) as User;
+			checkUser: async () => {
+				// lookup the user by username to see if already exists
+				const _user = (await store.dispatch('users/getGlobalUser', user.value.username)) as User;
+				console.log(_user);
```
drgrice1, line 95: Delete all `console.log`'s before committing.  A general practice you should get used to is to run `npm run build` or `quasar build` before committing.  `console.log`'s will break the production build.

drgrice1: Furthermore, the production build runs all linters with errors.  The development build does not.

package.json
pstaabp: added this as well.

src/components/admin/AddCourse.vue
pstaabp: done.

PR #35, opened 2021-09-03
Upgrade Quasar, Documentation and creation of right sidebar

pstaabp:
I apologize about putting in too much disparate stuff in one PR.  This has the following

- upgrades to quasar/other node modules
- improved POD documentation of the database modules
- creates a right sidebar in the UI to allow some other functionality.
- other bug fixes

I'll try to focus PRs better in the future


lib/DB/Schema/Result/PoolProblem.pm
```diff
@@ -3,10 +3,50 @@ use base qw/DBIx::Class::Core/;
...
+
+C<library_id>: a database id of the problem (foreign key)
+
+=item *
+
+C<problem_path>: alternatively, the path of the problem
+
+=back
+
+=head4
```
drgrice1, line 38: I think you are going to need to drop the `=head4` here.  Headers can't be inside an item list.  Either that or move the `=back` below above this line, but then the note is not at the same level.

In any case, this causes a POD error.

conf/course_defaults.yml
```diff
@@ -182,7 +185,7 @@
   subcategory: reduced_scoring
   doc: Value of work done in Reduced Scoring Period
   doc2: >
-    After the Reduced Scoring Date all additional work done by the student
+    After the Reduced Scoring Date all additional work done by the student`
```
drgrice1, line 188: Typo?

lib/DB/Schema/Result/Problem.pm
```diff
@@ -7,9 +7,62 @@ use warnings;
...
+=head4
+
+Note: a problem should have only one of a library_id, problem_path or problem_pool_id
+
+=cut
+
 sub valid_params {
 	return {
-		weight          => q{\d+},
+		weight          => q{[1-9]\d?},
```
drgrice1, line 65: I am not sure about this.  Why restrict to 2 digits?  In fact, I would like to see this allow decimals as well.

lib/DB/Schema/Result/ProblemSet/Quiz.pm
```diff
@@ -2,23 +2,75 @@ package DB::Schema::Result::ProblemSet::Quiz;
...
+
+=back
+
+=cut
+
 sub valid_params {
 	return {
 		timed => q{^[01]$},
-		time_length => q{\d+},
+		quiz_length => q{\d+},
```
drgrice1, line 63: Maybe quiz_duration would be better?  quiz_length doesn't seem entirely clear as to what length it is about (without reading the POD of course).

src/layouts/MainLayout.vue
```diff
@@ -1,33 +1,68 @@
...
+			<problem-set-list v-if="show_problem_sets"/>
+			<user-list v-if="show_users" />
+		</q-drawer>
+
 	</q-layout>
 </template>
 
 <script lang="ts">
-import { defineComponent, ref } from 'vue';
+import { defineComponent, ref, watch, Ref, computed } from 'vue';
```
drgrice1, line 22: `computed` is defined but not used.

lib/WeBWorK3/Controller/ProblemSet.pm
```diff
@@ -31,6 +31,7 @@ sub getProblemSets {
 		$set->{set_visible} = $set->{set_visible} ? true : false;
 	}
 	$self->render(json => \@problem_sets);
+	return;
```
drgrice1, line 34: Why the `return` statement here?  A return statement at the end of a sub that doesn't return anything is generally not needed.  I actually have a branch that removes all of the others in this file.  That is just clutter since the return values of these methods are not used for anything.

drgrice1 (approved):
This generally looks good.  There are a few minor issues though.

There are also some things that we probably need to discuss about the implementation.

lib/DB/Schema/Result/Problem.pm
pstaabp: Seems reasonable.

lib/DB/Schema/Result/ProblemSet/Quiz.pm
pstaabp: good point.

lib/WeBWorK3/Controller/ProblemSet.pm
pstaabp: This is something that comes up on perlcritic when I run it (but I forget what level).  Perhaps if we agree that we'll abide to perlcritic level 5 (I think that's the lowest), then I'll remove all of the unnecessary returns.

drgrice1: Yeah, I thought that might be the case.  We can leave it I guess.

lib/DB/Schema/Result/PoolProblem.pm
pstaabp: fixed.

conf/course_defaults.yml
pstaabp: yes.

lib/DB/Schema/Result/Problem.pm
pstaabp: made it positive integers or decimals.

lib/DB/Schema/Result/ProblemSet/Quiz.pm
pstaabp: fixed.

src/layouts/MainLayout.vue
pstaabp: fixed.

lib/WeBWorK3/Controller/ProblemSet.pm
pstaabp: Seems to be need for level 4 of perlcritic.  I guess we should just determine what level we want.

pstaabp:
Pushed up some changes.

drgrice1 (approved):

drgrice1:
I think this is ready, so I will merge it.

PR #36, opened 2021-09-05
Use the standalone renderer to render a problem

pstaabp:
This uses the standalone renderer to develop a sample example of a Problem Viewer.  To test this

1. install the renderer.  Instructions are included in the README.md file.
2. The Problem Viewer shows some sample problems (just picks 1 of 3 at random).  No submission or checking is tested yet.


README.md
```diff
@@ -60,6 +60,24 @@ are assuming terminal/shell commands.
 5. If you get the message `Web application available at http://127.0.0.1:3000` then mojolicious is running and waiting
    for any requests.
 
+### Getting the standalone renderer code running
+
+1. Clone the repository with `git clone --recursive https://github.com/drdrew42/renderer`
```
drgrice1, line 65: I recommend that you point this to https://github.com/openwebwork/renderer.  At this point those are the same in terms of the code.  I have been ensuring that.  As such the code still references Andrew's branches, but I will change that soon.

drgrice1:
This isn't working for me.  I signed in as Lisa, and clicked the load problem button, and nothing happens.  The standalone renderer instance gives several errors.  The main one that seems to make it fail is "Problem created from  with no random seed."

README.md
pstaabp: True.  I was using the instructions on @drdrew42 renderer site.

pstaabp:
Oops.  There's a one-line change for the renderer that we need.  I posted the repository at `https://github.com/pstaabp/renderer/tree/ww3-problem-fix`.

Also, the math output for the renderer is in <script> tags and I'm not sure how to get MathJax to interpret these.  I did a regex replace on these, but I remember that we might run into some issues on some library problems.  It also appears that the renderer always sends back an entire html page as a problem (I think the intention is to run in an iframe).  Again, I strip the problem out of the code. 

Mainly I was using this as a test that we can get problems rendered, but obviously there's a lot of work required.

drgrice1:
Yeah, we won't want to do what you are doing with the script tags.  Instead we will need to load the MathJax configuration the same way that webwork2 does.

drgrice1:
Looking at your branch I see that the issue is what I suspected.  The renderer expects form data, and axios is posting json data.  Why not just post form data?  Axios can do that I believe.

pstaabp:
That's a good idea.  That way I don't have to change the renderer code.  I'll take a look.

drgrice1:
Here is a quick patch that switches to using formData instead of json:
```patch
 	setup(props){
 		const html: Ref<string> = ref('');
 		const _file: Ref<string> = ref(props.file);
+
+		const formData = new FormData();
+		formData.append('problemSeed', '1');
+		formData.append('sourceFilePath', _file.value);
+		formData.append('outputFormat', 'static');
+
 		async function loadProblem() {
-			const response = await axios.post('/renderer/render-api',
-				{
-					problemSeed: 1,
-					sourceFilePath: _file.value,
-					outputFormat: 'static'
-				});
+			const response = await axios.post('/renderer/render-api', formData,
+				{ headers: { 'Content-Type': 'multipart/form-data' } });
 			console.log(response);
 
 			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```

pstaabp:
@drgrice1 This looks good.  However, just noticed that one of the problems is reporting a `cannot find macro` error.  Maybe I'm missing some configuration setting.

src/components/common/Problem.vue
```diff
@@ -0,0 +1,72 @@
...
+	},
+	setup(props){
+		const html: Ref<string> = ref('');
+		const _file: Ref<string> = ref(props.file);
+
+ 		async function loadProblem() {
+			const formData = new FormData();
+			formData.append('problemSeed', '1');
+			formData.append('sourceFilePath', _file.value);
+			formData.append('outputFormat', 'static');
```
drdrew42, line 33: Not an issue here, but generally we will want to use `.set()` rather than `.append()` in that `set` will replace an existing key instead of `append` which will duplicate keys... it's just 'safer'

The exception is when WeBWorK problems use checkboxes, in which case it is expected that the same key might be repeated in the case that multiple boxes are checked. (I am 99% that is the only exception...)

drdrew42 (approved):
I'm able to get all three problems to render

README.md
```diff
@@ -60,6 +60,24 @@ are assuming terminal/shell commands.
...
+
+2. copy `render_app.conf.dist` to `render_app.conf` and make any desired modifications including changing the ports
+from 3000 to 3001
```
drdrew42, line 68: Because quasar is proxying the renderer at `/renderer`, the following configurations are necessary in `render_app.conf`:
* SITE_HOST => 'http://localhost:3000' (point at quasar)
* baseURL => '/renderer' (proxy endpoint for SITE_HOST)

Assuming we are playing MITM, users should never have access to the renderer directly.

src/components/common/Problem.vue
drgrice1: Yeah, set is probably safer than append here.  I quickly hacked this together for Peter this morning.  I didn't check the FormData api.

README.md
drgrice1: Since the dev server is rewriting /renderer to root, it isn't strictly necessary to set the baseURL setting of the renderer.  Although it would probably be better to do it that way rather than using a rewrite.

drgrice1: In terms of actual deployment, this will all need to be done differently in any case.  The webpack dev server is not meant for production use.

drgrice1:
@psaapb:  Make sure that you remove the console.log's before this is merged.  Remember that those break the build.

drgrice1:
#pstaabp:  Can you test my pull request to this pull request?  That fixes the mathjax configuration to deal with the script tags correctly.  It also fixes some of the other things discussed here.

Then we can merge this in.

README.md
pstaabp: On `SITE_HOST`, I was using :3001 for this, however I think that was mainly for pulling js libraries when the renderer was sending a full HTML page.  I have stripped it down to only the problem within the `<form>` tag.   

Also, I've kept the `baseURL` as `/` but maybe because again pulling out only the problem.

pstaabp:
Just pushed an update which allows mathjax to render math in the `<script>` tags and updated some documentation.

drgrice1:
I think I will go ahead and merge this, and then I will put my pull request in directly.  I think @pstaapb is busy with school starting right now.  Then we can continue with the other things that depend on this.

@pstaabp:  Does that sound good?

drgrice1:
By putting my pull request in directly, I mean I create a pull request to the main repository.

drgrice1:
> @drgrice1 This looks good. However, just noticed that one of the problems is reporting a `cannot find macro` error. Maybe I'm missing some configuration setting.

I am getting this too now.  I was not getting this before. I am not sure what has changed.

drgrice1:
I forgot one step that I usually do when I start the renderer.  This is probably the reason you get the macro not found error also @pstaapb.  Make sure that the WEBWORK_ROOT environment variable is unset.  Of course on my system that is set up for webwork2 usage, that is usually set.  Yours is probably that way as well.  If you `unset WEBWORK_ROOT` the macro file is found.

PR #37, opened 2021-09-07
Fixing configuration of problem sets

drdrew42:
Draft PR -- will open after merging #36 

### [store/common.ts] parseBoolean

Responses from the Mojo backend are already parsed as boolean, causing `set_visible` to be interpreted as 'false' regardless of their actual state. `parseBoolean` now accepts and returns booleans, and properly interprets 'true', 'false', undefined, as well as 0/1 as either strings or ints.

'new' problem sets (HWset, Quiz, etc) have all param keys present, despite some values being undefined. This causes issues with `store/modules/problem_sets.ts` `updateSet` because the requested changes and the response fail to be equal due to the presence of keys with undefined values. Using lodash `pickBy`, the params object is filtered for keys with defined values. This satisfies `updateSet` and ensures that the store is updated when the db is updated.

### [DB/Schema/Result/ProblemSet/ReviewSet.pm]

Needed to inherit from `DB::WithParams` and `DB::WithDates` or else updates failed with missing methods `validParams` and `validDates`.

### [DB/WithParams.pm] and [DB/WithParams.pm]

`valid_params` and `required_params` were not being loaded if they already had a value -- but this means that updating sets only worked so long as you stuck with the same type of set.  e.g. Modifying a HWSet would function properly, but continuing on and modifying a ReviewSet would fail because the valid/required params were based on HWSet and not ReviewSet.

Removing the conditional on the definition of `valid_params` and `required_params` caused "variable ... is not available" warnings, so they have been scoped as `our` -- I do not know if this is an ideal solution.

### [t/db/006_quizzes.t]

The various tests for `addProblemSet` need to specify `set_type => 'QUIZ'`, or else they're testing the default `HWSet`...


lib/WeBWorK3/Mojolicious.pm
```diff
@@ -16,7 +16,7 @@ our $exception_handler = sub {
 			$next->();
 		} catch {
 			my $output = {exception => ref($_)};
-			$output->{message} = $_->message if ($_ && ref($_) eq 'Mojo::Exception' or ref($_) =~ /^DB::Exception/x);
+			$output->{message} = $_->message if ($_ and ref($_) eq 'Mojo::Exception' or ref($_) =~ /^DB::Exception/x);
```
drgrice1, line 19: Note that "a && b or c" is logically equivalent to "a and b or c" (these are also logically equivalent to "a && b || c").  So in terms of what is desired, this doesn't change anything.  Of course perl critic complains about this with the policy ValuesAndExpressions::ProhibitMixedBooleanOperators set at its defaults and severity level 4, which is probably why you made this change.  (Further note that "a and b || c" is logically different than "a and b or c" and the others.)

However, both before and after this change this is logically incorrect.  Parentheses are necessary in this situation around the "or" part.  With the way it was before and with this change if `$_` is falsy, then the last condition will still be checked, and that should not be the case here.  (In fact "a and b || c" does work correctly here without parentheses, although it breaks the perlcritic policy.)

Furthermore, you made the wrong change to satisfy the no mixed boolean operator policy in my opinion.  You changed the `&&` to `and`, and you should have changed `or` to `||`.  Generally speaking always prefer `&&` and `||` to `and` and `or` unless it is one of the very few specific situations where the latter should be used (like for control flow).

So in short the correct (as well as opinionated) change should be:
```perl
$output->{message} = $_->message if ($_ && (ref($_) eq 'Mojo::Exception' || ref($_) =~ /^DB::Exception/x));
```

pstaabp:
Overall, looks good.  I'll test after merge of #36

lib/WeBWorK3/Mojolicious.pm
drdrew42: I think these are changes from #36 -- I don't recall making this change.

drdrew42: Also, agreed -- && > 'and', || > 'or'

drgrice1: I see. That is what I get for commenting before that was merged. I can't sort out what is this pull request and what is that.

drdrew42: I think I covered all the 'changed files' in the description -- but yeah, it will be much easier to review once the previous PR gets merged...

drgrice1: Oh, I was supposed to read that!

drgrice1:
@drdrew42:  If you merge the main branch into this one, then the lint workflow will run again, and will pass.  The main branch now has a change to the copy/paste detector configuration (jscpd) that disables it for Perl and TypeScript files, and that is what is failing here.

drdrew42:
Now there's a new linting issue...

From t/db/006_quizzes.t:
`Code not contained in explicit package at line 4, column 1.  Violates encapsulation.`

drgrice1:
Odd. I am not getting that when I run perlcritic locally on the file.

That can be fixed by adding `#!/usr/bin/env perl` to the first line of the file.  That will probably be needed in all of the test files then.

drgrice1:
Wait, I am getting several warnings on the `.t` files.  I was testing on the `main` branch, and that doesn't have the .perlcriticrc file.  I didn't realize you merged my `api-cleanup` branch into this.  I will fix this in that branch.  I didn't think to test the `.t` files.  I only checked the `.pm` and `.pl` files.

drgrice1:
Yay!  Lint clean!

drgrice1:
👍

PR #42, opened 2021-09-12
Render problem without iframe

drdrew42:
## Getting Started
A few "minor" things first (@drgrice1 make sure to read this part 😜 ):
* in this PR, we are not rewriting request URLs when proxying to the renderer
  * pull the latest drdrew42/renderer and switch to the `feature/no-iframe` branch
  * make sure to update your render_app.conf accordingly (`baseURL => '/renderer'` and `SITE_HOST => 'http://localhost:8080'`)
  * inside the renderer, point the PG submodule to drdrew42/pg branch: `renderer/separate-css-js`
* I changed a few problem paths so that our test problems have js/css

## Known Issues
* No order is specified when loading js/css resources
  * As such, GeoGebra may be embedded, but is not initialized
  * Similarly, MathQuill may fail because jQuery is not yet loaded
* In scaffolded problems, the app dies if you toggle a scaffold
  * likely collision between bootstrap + quasar?
* console.logs errywhere


drdrew42:
## Changes
* html is now updated _after_ awaiting all injected css/js resources
* pull PG branch `renderer/separate-css-js` again, no need to inject mathjax twice

## Good News
* it seems that MathJax works across multiple renders
* 'normal' answer blanks are hidden
* GeoGebra applet displays the first time the problem is randomly called

## Bad News
* MathQuill is not being loaded?
* GeoGebra applet does not display on subsequent calls
* GeoGebra applet is never properly initialized
* Scaffold + knowl clicks redirect to login page (unsupported route, most likely)

src/components/common/Problem.vue
```diff
@@ -29,34 +30,107 @@ export default defineComponent({
...
+		// modified from https://github.com/tserkov/vue-plugin-load-script
+		async function loadScript(src: string) {
+			console.log('loading scripts...', src);
+			// eslint-disable-line no-param-reassign
+			return new Promise<void>(function (resolve, reject) {
+				let shouldAppend = false;
+				let el: HTMLScriptElement | HTMLLinkElement | null;
+				create_el: if (/\.js(?:\??[0-9a-zA-Z=]*)$/.exec(src)) {
+					el = document.querySelector('script[src="' + src + '"]');
+					if (el && el.hasAttribute('data-loaded')) {
```
drgrice1, line 42: This can be simplified using optional chaining.  Just change it to `if (el?.hasAttribute('data-loaded')) {`.

src/components/common/Problem.vue
```diff
@@ -29,34 +30,107 @@ export default defineComponent({
...
+						resolve();
+						console.log(`${src} seems to have already been loaded!`);
+						return;
+					} else if (el) {
+						break create_el;
+					}
+
+					el = document.createElement('script');
+					el.type = 'text/javascript';
+					el.async = true;
```
drgrice1, line 52: This should be changed to `el.defer = true`.  Asynchronous loading does not guarantee order of loading, and defer does.

src/components/common/Problem.vue
```diff
@@ -29,34 +30,107 @@ export default defineComponent({
...
+					return;
+				}
+
+				el.onerror = reject;
+				el.onabort = reject;
+				el.onload = function loadScriptHandler() {
+					console.log(`[onLoad] ${src} successfully loaded!`);
+					el?.setAttribute('data-loaded', 'true');
+					resolve();
+				};
```
drgrice1, line 82: These events should be added with addEventListener instead of assigning to the on... attributes.  So, for example, change to `el.addEventListener('error', reject);`.

src/components/common/Problem.vue
```diff
@@ -29,34 +30,107 @@ export default defineComponent({
...
-			// Extract only the form element.
-			const match = /<form(.*)>(.*)<\/form>/s.exec(value);
-			if (match) html.value = match[0];
+			await Promise.all(css.map(
+				async (jsSource) => {
+					await loadScript(jsSource).
+						// then(() => console.log(`loaded ${jsSource}`)).
+						catch(() => console.error(`Could not load ${jsSource}`));
+				}
+			));
```
drgrice1, line 125: We are not going to be able to use Promise.all if we want to guarantee the order that the scripts and styles are loaded.  This could be done with an await call inside a for loop, but that is generally bad practice.  Although, there really is no reason to load each script or css file in a separate async function call.

Also, the css files should definitely be loaded before the javascript files.

drgrice1 (commented):
Note my comments in the code.

Furthermore, the ww3_format will need more changes to make this work well.  The Bootstrap css classes used will not work.  Note that span12 is a Bootstrap 2 css class not used by recent versions of Bootstrap, and we will certainly not use Bootstrap 2 anymore, even if we do use Bootstrap.  This is true for some of the other classes as well.  Furthermore some of the css classes that are still valid for recent versions of Bootstrap conflict with quasar classes.  Although in general all of the added html will need to be removed, and the client will need to build that from the basic elements.  That will be the only way that this can be made to look good in the vue/quasar environment, unless we use run time template compilation and switch to a vue template.

Of course all console.log's will need to be removed before committing, but this is still a draft.  The build will break if those are present.

src/components/common/Problem.vue
drgrice1: I should add that I doubt defer will actually have any effect in this situation since these scripts are being dynamically injected after the DOM has already loaded in any case.  On the other hand async will have the negative implications I mentioned.

src/components/common/Problem.vue
```diff
@@ -29,34 +30,107 @@ export default defineComponent({
...
 
+			console.log('All js and css have been loaded...');
+			html.value = value;
 			/* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call */
 			if (window.MathJax && typeof window.MathJax.typesetPromise == 'function') {
 				// eslint-disable-next-line @typescript-eslint/no-unsafe-return
 				window.MathJax.startup.promise.then(() => window.MathJax.typesetPromise([renderDiv.value]));
 			}
-			/* eslint-enable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call */
```
drgrice1, line 59: This line should be added back.  Otherwise these rules are disabled for the rest of the file instead of only for that code block as was the intent.

src/components/common/Problem.vue
```diff
@@ -29,34 +30,107 @@ export default defineComponent({
 		const _file: Ref<string> = ref(props.file);
 		const renderDiv = ref<HTMLElement>();
 
+		// modified from https://github.com/tserkov/vue-plugin-load-script
+		async function loadScript(src: string) {
+			console.log('loading scripts...', src);
+			// eslint-disable-line no-param-reassign
```
drgrice1, line 36: Note that `eslint-disable-line` only works at the end of the line and only affects the line it is on.

In any case, this is not needed.  Eslint does not warn about this.

drgrice1:
I don't think that there is a need for the load_css and load_js subroutines that you added to PG.pl in your pg branch.  Webwork3 will add the neccessary javascript and css.  Many of those that you are loading aren't going to work anymore for webwork3.

Also, mathquill and the mqeditor should not be loaded unless the course setting for that is enabled.  Of course that needs to be set up yet, and you probably are doing that for testing purposes.

pstaabp:
Finally getting around to testing this.  I'm a bit confused on which ports are used.  By default quasar/node will run on 8080.  The mojolicious for webwork3 is on 3000.  Because of the settings for the renderer, should this be on 3001 still or 8081?

drgrice1:
All should listen on the same ports as before. The SITE_HOST settings of the renderer is the requesting site, and so should be port 8081.

pstaabp:
I'm getting the following errors in the console where I'm running `quasar dev`

```
Issues checking in progress...
Error: Debug Failure. Unhandled Diagnostic: 2626
Error: Debug Failure. Unhandled Diagnostic: 2626
    at reportIncompatibleStack (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:59271:45)
    at reportRelationError (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:59314:21)
    at reportErrorResults (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:59580:25)
    at isRelatedTo (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:59539:17)
    at typeArgumentsRelatedTo (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:59842:39)
    at relateVariances (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:60394:34)
    at structuredTypeRelatedToWorker (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:60330:46)
    at structuredTypeRelatedTo (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:59978:30)
    at recursiveTypeRelatedTo (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:59950:64)
    at isRelatedTo (/opt/webwork/webwork3/node_modules/typescript/lib/typescript.js:59490:34)
```

Also, trying to load a problem in the Problem View, I'm getting
```
POST http://localhost:8080/renderer/render-api 404 (Not Found)
```

Clearly, I don't have the settings right.

drgrice1:
This is not caused by the settings. This is caused by the presence of the dist folder. Delete that for now. This is really caused by the presence of the MathJax files in the dist folder that I have now installing the webpack copy plugin.  I am not sure why eslint is trying to process those files. I do the exact same thing on another project in this doesn't happen.

pstaabp:
> This is caused by the presence of the dist folder. Delete that for now.

That fixed the 2626 error.     I was on the wrong webwork3 branch.  It's running now.  Resume testing...

src/components/common/Problem.vue
drdrew42: I did this to see if there was any difference in waiting for all the js/css to be loaded before inserting the renderedHTML

drdrew42:
Moving ahead with this:
1. should the ww3 renderer format be stripped down to the form tags -- or remove those as well?
2. should the 'common' js/css dependencies simply be included in this page instead of relying on this insertion technique?

drgrice1:
The css and javascript loaded by the current webwork2 template will need to be handled in this page, rather than injected dynamically.  The Bootstrap css will take special care, and that will only work if we use a newer version of Bootstrap.  Basically we need to deep scope the css for the vue template, or it will mess up the entire SPA.   That is not a real possibility with Bootstrap 2 as it does not provide the sass files needed for this.

There are a lot of complications to getting the non-iframe approach to work.  I think it might be easiest to talk about it on Thursday, but here are a few things I have thought of.

Much of the current webwork2 css and javascript will need to be rewritten to work with this.  Anything that uses the window load or DOMContentLoaded events will not work.   There are other things that will need to be changed also.  Currently some javascript needs to be loaded before the problem html is loaded, and there is some that needs to be loaded after.  We will need to rewrite those to be uniform. 

There also may be a problem with script and style tags that are contained in the the problem text itself.  Some problems add this.  That will take some care as well.

This all may be quite a bit, and so perhaps it would be better to go with the iframe approach (at least for now) as you were suggesting.

pstaabp:
I didn't get the webwork2 directory set up correctly, so am getting errors loading any js or css out of the htdocs directory.  It seems that eventually these should live on the PG side of things. 

Also, the Base64 stuff seems to be built into almost any browser these days with the `atob` and `btoa` functions.  Perhaps we can drop reliance on `Base64.js` assuming these are equivalent.

pstaabp:
I have `SITE_HOST`  set to `http://localhost:8081` and `baseURL => '/renderer',` but getting errors loading anything out of `webwork2_files`.  However, if I request: `http://localhost:8080/renderer/webwork2_files/js/apps/Base64/Base64.js` this is working.  Are others seeing this problem?  Or do I not have setting correct.

pstaabp:
> Much of the current webwork2 css and javascript will need to be rewritten to work with this. Anything that uses the window load or DOMContentLoaded events will not work. There are other things that will need to be changed also. Currently some javascript needs to be loaded before the problem html is loaded, and there is some that needs to be loaded after. We will need to rewrite those to be uniform.

Perhaps an approach to some of the problems might be to rewrite the backend macros instead of deal with the js files.  I'm thinking of `Scaffold.pl` for example, which  has a few commands written for bootstrap (lines 616--625).  We could rewrite those for quasar and the `q-expansion-item` framework.  We could then output a different template when `ww3-output` is present.   No js would be needed then.   I don't think there are a lot of PG pieces that use js.

drgrice1:
> Also, the Base64 stuff seems to be built into almost any browser these days with the `atob` and `btoa` functions. Perhaps we can drop reliance on `Base64.js` assuming these are equivalent.

Yeah, that is somethings that needs to be done.  I have a rewrite of the applet support javascript that will do that.

drgrice1:
> Perhaps an approach to some of the problems might be to rewrite the backend macros instead of deal with the js files. I'm thinking of `Scaffold.pl` for example, which has a few commands written for bootstrap (lines 616--625). We could rewrite those for quasar and the `q-expansion-item` framework. We could then output a different template when `ww3-output` is present. No js would be needed then. I don't think there are a lot of PG pieces that use js.

I don't think that is the correct approach.  The javascript files for pg should be on the pg side as we have discussed already, and those should be independent of the frontend library that is used.  I don't agree with having different output from pg for a specific frontend.  Also, I think there are a lot more instances of javascript usage than you say, and there will only be more usage of this as time goes on.  At this point the scaffold macro, the applet macro, the graphtool macro, and the draggable proof macro use javascript, and there are probably others too (I know there is some matrix thing that Mike wrote, but I can't find it right now).

drdrew42:
> > Also, the Base64 stuff seems to be built into almost any browser these days with the `atob` and `btoa` functions. Perhaps we can drop reliance on `Base64.js` assuming these are equivalent.
> 
> Yeah, that is somethings that needs to be done. I have a rewrite of the applet support javascript that will do that.

`atob` and `btoa` do not properly support UTF-8. I was doing this with the standalone renderer until Tani suggested changes.

drdrew42:
> I have `SITE_HOST` set to `http://localhost:8081` and `baseURL => '/renderer',` but getting errors loading anything out of `webwork2_files`. However, if I request: `http://localhost:8080/renderer/webwork2_files/js/apps/Base64/Base64.js` this is working. Are others seeing this problem? Or do I not have setting correct.

SITE_HOST should be port 8080 -- the request should pass through the proxy (in production, this will be handled by a load balancer or nginx etc)

drgrice1:
Yes, atob and btoa do not support UTF-8.  However, we are only using this in javascript in 2 places.  The first is for saving state for Geogebra applets, and the second is for decoding knowls.

For Geogebra applet state I don't think that UTF-8 support is needed and atob and btoa should work fine.  This could also be reworked to either use Geogebra's built in base64 encoding and decoding (setBase64 and getBase64) or to drop base64 usage and instead use json (getFileJSON and setFileJSON).

Knowl's shouldn't use base64 encoding at all.  I have a rewrite of the knowl javascript and usage in webwork that removes that.

drgrice1:
> All should listen on the same ports as before. The SITE_HOST settings of the renderer is the requesting site, and so should be port 8081.

My bad.  That should have been 8080.

src/components/common/Problem.vue
```diff
@@ -29,28 +30,100 @@ export default defineComponent({
...
+					}
+
+					el = document.createElement('script');
+					el.type = 'text/javascript';
+					el.defer = true;
+					el.src = src;
+					shouldAppend = true;
+				} else if (/\.css(?:\??[0-9a-zA-Z=]*)$/.exec(src)) {
+					el = document.querySelector('link[href="' + src + '"]');
+					if (el && el.hasAttribute('data-loaded')) {
```
drgrice1, line 56: I guess I should have pointed out this instance that is the same as line 41.  I assumed you would catch both of them.  Bad assumption on my part.

drdrew42: Apparently I got too excited to check off all the feedback... 😂

drdrew42:
Closed in favor of #47

PR #43, opened 2021-09-13
User manager changes

pstaabp:
There's a number of changes to this
1. refactoring the code to split some of the utilities associated with users, problem sets, course settings, courses. Some validation and parsing code has been added. 
2. Update the classlist manager
3. Add a "Add Users From File" option to the classlist manager.  The file should be a CSV and the headers can be selected.  

- Note: more error checking needs to be handled on this.  
- There is a sample CSV in `t/db/sample_data/mlb_list.csv` that I used to pull in a lot of data (MLB home run leaders), but others should be checked too.
- Before being pulled in (or maybe a new PR), I plan to handle more validation errors in a robust way.

A new package has been added (`papaparse`), so a `npm install` is needed before testing.


lib/WeBWorK3/Hooks.pm
```diff
@@ -19,6 +20,7 @@ our $exception_handler = sub {
 			$output->{message} = $_->message
 				if (ref($_) && (ref($_) eq 'Mojo::Exception' || ref($_) =~ /^DB::Exception/x));
 			$output->{message} = $_ if ($_ && ref($_) eq 'DBIx::Class::Exception');
+			$log->error($output->{message});
```
drgrice1, line 23: There is no need to `use Mojo::Log` in this file and create a new instance of that object.  Just use the already existing application log instance.  For that change this to
`$c->app->log->error($output->{message});`

t/db/sample_data/mlb_list.csv
```diff
@@ -0,0 +1,51 @@
+HR,First_Name,Last_Name,Username,Email,Section
```
drgrice1, line 1: What is the HR column?

drgrice1 (commented):
Here are a few initial observations.   See my notes on the code.

Also, exception handling needs to be rethought for the api.  Exception handling should only be used for unexpected situations that you don't really intend to handle properly.  They should not be used in the broad blanket form that they are currently set up with.  For example, if the database is queried for a user that does not exist, an exception should not be thrown.  It is expected that users will sign in that are not valid users, and this should be handled properly.  In this particular example this prevents the messages from being rendered and sent to the client.  Instead the blanket handler sends the exception name. 
 Database queries for non existent data in general is expected and should be handled gracefully.

lib/WeBWorK3/Hooks.pm
drdrew42: I think you can shorten that to `$c->log->error(...)`

drgrice1: Yep.  It looks like you can.

t/db/sample_data/mlb_list.csv
pstaabp: It is the number of home runs. I have been using it for the student id. 

I'm thinking of making this very robust that you can load in any csv file and handle it.  The "header rows" toggle guesses what the columns are, but you can change those to whatever.

pstaabp:
About exception handling, I agree that much more thought needs to occur.  

My hope is that the UI is designed well enough to catch everything before sending things to the api/backend.  However, the way I've designed the DB is to have it throw exceptions that make some sense (like `UserNotFound`), so perhaps they can be parsed if needed.

pstaabp:
Added some extra error checking of loaded users with some UI to handle errors.  Also, there is a file `t/db/sample_data/bad_user_list.csv` that has some invalid fields.

drgrice1 (commented):
In addition to my comments in the code, there are a few minor issues.

I can't seem to add a user via either from a file or with the manual method.

Also, if I delete a user, a toast pops up that says the user is deleted, yet the user is still appears in the list.  If you look in the database, you see that the user has been remove from the course although not deleted entirely.  It seems that the user should be deleted entirely if not in any courses.

Also, the user can delete him/herself from the course.

drgrice1:
@pstaabp:  I put a pull request in to your branch for this pull request to switch from sass to scss, and deal with a few of the other minor things I observe in the review.

src/components/admin/AddCourse.vue
```diff
@@ -60,7 +60,8 @@ import { defineComponent, ref, Ref } from 'vue';
 import { useQuasar } from 'quasar';
 
 import { useStore } from '../../store';
-import { newCourse, newCourseUser, newUser } from '../../store/common';
+import { newCourseUser, newUser } from '../../store/utils/users';
+import { newCourse } from 'src/store/utils/courses';
```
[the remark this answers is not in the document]
pstaabp: I moved a bunch a files around and my editor refactored the code to reflect this.  Will clean up.

lib/WeBWorK3/Hooks.pm
```diff
@@ -3,11 +3,10 @@ use warnings;
 use strict;
 
 use Try::Tiny;
+use Mojo::Log;
```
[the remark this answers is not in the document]
pstaabp: fixed.

t/db/sample_data/mlb_list.csv
drdrew42: If I toggle the header row on before loading the csv, the file is not parsed as though it has a header row. If I then turn the header row off, the data vanishes. I have to load the file without header row on, and once the data is loaded, the header row will be parsed correctly.

Hiding the header row toggle until the data is loaded would solve this issue, and simplify the import workflow.

src/store/utils/users.ts
```diff
@@ -0,0 +1,125 @@
...
+	return true;
+}
+
+export function parseCourseUser(params: Dictionary<string|number>): CourseUser {
+	const course_user = newCourseUser();
+	const user_fields = Object.keys(course_user);
+	// check that the required fields are present in the params
+	const common_fields = intersection(required_course_user_params, Object.keys(params));
+	if (!isEqual(common_fields, required_course_user_params)) {
+		const diff = difference(required_user_params, common_fields);
```
drdrew42, line 92:
```suggestion
		const diff = difference(required_course_user_params, common_fields);
```

drdrew42: Comparing the wrong array here... looks like it didn't get updated when copy-pasted from `parseUser`

src/components/instructor/AddUsersFromFile.vue
```diff
@@ -1,23 +1,325 @@
...
+				let course_user: Dictionary<string|number> = {};
+
+				try {
+					user = pick(mapValues(user_param_map.value, (obj) => params[obj]), Object.keys(newUser()));
+					users_to_add.value.push(parseUser(user));
+
+					course_user = pick(mapValues(user_param_map.value, (obj) => params[obj]),
+						Object.keys(newCourseUser()));
+
+					const parsed_course_user = parseCourseUser(course_user);
```
drdrew42, line 181: before parsing `course_user`, we need `course_user.role = course_user.role ?? 'student'` or else `parseCourseUser` will fail.

Alternatively, `parseCourseUser` should not throw an error with a missing user.role, and instead provide 'student' as a default.

pstaabp: Agree with this.  I've been thinking of a better way to handle all of the parsing.

t/db/sample_data/mlb_list.csv
pstaabp: I disabled the header row toggle until the file was loaded.

lib/DB/Schema/ResultSet/User.pm
```diff
@@ -105,10 +105,13 @@ sub addGlobalUser {
 	my ( $self, $user_params, $as_result_set ) = @_;
 	DB::Exception::ParametersNeeded->throw( message => "The parameters must include username" )
 		unless defined( $user_params->{username} );
+	my $params = clone($user_params);
+	# remove the user_id if defined and equal to zero.
+	if ( defined ($params->{user_id}) && $params->{user_id} == 0) {
+		delete $params->{user_id};
+	}
```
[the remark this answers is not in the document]
pstaabp: The `user_id` field is autogenerated.  On the client side when we make a new user, I assume the `user_id` to be 0.  However, on the DB side, if it is zero, it's a new user, so delete it and let the DB generate the `user_id`. 

Alternatively (and as I write this it makes more sense), to not assign a `user_id` on the client side.  I think I had reason for not doing that, but I can't remember. 

If we keep with the new user has `user_id` of 0, I'll add a comment to the code.

src/store/utils/users.ts
pstaabp: Yep.  And this is another reason for developing a Model class that does the validating and let's everything inherit from it.

pstaabp:
Pulled in your PR @drgrice1.  Checked in some fixes to the code.  However, I'm wrestling with how to handle course users.  Check out my thinking on Discussion #52.

lib/DB/Schema/ResultSet/User.pm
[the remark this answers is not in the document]
drgrice1: Right.  I am confusing the `user_id` with the `username`.  With webwork2 there was no `username`, and the `user_id` played that role.

I agree that the `user_id` probably should not be allowed to be set from the client, and should always be auto generated.  

There may be some issues with these database ids when it comes to course archival and restoration though.  If a course is archived, do we save these database ids, and then also restore them?  That could be problematic if during the time that the course is archived another user is created that has the same id.   If the course is archived and restored and the ids change, then the browser url for pages the user accesses will change, so that could be a problem if these ids are not saved and restored.  We may need to rethink using route params in the way that you have the vue router set up.

The route param issue has been on my mind for a while.  This may also affect things like LTI authentication.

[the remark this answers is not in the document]
pstaabp: About the archiving, I don't think we can keep the database ids, because there may be clashes with existing one.  I think if we import a course (for example), we add the users as new users and let the DB regenerate new ids. 

Say more about the route param issue?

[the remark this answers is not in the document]
drgrice1: What you are talking about is exactly the route param issue.  If a course is archived and then restored with this plan, then the urls for a user will all possibly have changed due to the course_id and user_id having changed.  For example, before archival the route may have been `/webwork3/courses/4/instructor`, and after restoring at some later time it will now be perhaps `/webwork3/courses/8/instructor` assuming several courses were added in the mean time.  This has implications for linking into webwork externally that we are going to need to figure out.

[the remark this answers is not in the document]
pstaabp: I see what you mean.  Alternative to the course_id we could use the course name in the route.  That would be more robust.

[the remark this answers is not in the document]
drgrice1: Perhaps, but that may not work so well either because with the new database scheme course names can have characters that might be problematic.

src/components/admin/AddCourse.vue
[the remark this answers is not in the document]
drgrice1: Maybe it would be nice to add our own webpack alias for `store` so that we don't always have to pass through `src` for that.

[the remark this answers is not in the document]
drgrice1: This is good for now though.  That is something we can add later.

pstaabp:
I think this is ready to be tested again.  Both adding students manually and via file should be tested as well as updating and deleting.

drdrew42:
Only just getting started with testing, here's what I see so far:
* blank error pop-up when adding users from file (when they already exist) -- console error shows text, but UI error message is empty.
* bad_users -- validation succeeds? (I am allowed to select the rows.) -- console error shows text, but UI error message is empty.
* console errors when canceling out of file selection `Cannot read properties of undefined (reading 'name')` AddUsersFromFile.vue L13
* editing user from student -> TA creates duplicate entry in user display table (user appears as student & TA, selecting one row selects both), duplication seems to have *replaced* one existing user (bbonds went missing when the duplication occurred)
* deleting the duplicated user removed all users that were previously added from file and caused an error `no such column: me.course_user_id` User.pm line 438
* attempting to re-add the (unintentionally) deleted users from file caused errors (username: xxx is already in the course: xxx) -- but those users are not displayed unless I log out and log back in.

pstaabp:
@drdrew42 I'll work on bug fixes.  So wait until further testing.

PR #44, opened 2021-09-13
Use iframe to render problems

drdrew42:
## Notes
* adds `iframe-resizer` -- so make sure to `npm install`
* make sure to put the renderer back to `master` branch (and pull for minor update)

~~Definitely renders more cleanly, though I'm seeing:
`jQuery.Deferred exception: $(...).tooltip is not a function TypeError: $(...).tooltip is not a function`
and
`Uncaught TypeError: $(...).tooltip is not a function`~~

_Edit:_ console errors were caused by the no-iframe pg branch... make sure to switch back!


src/components/common/Problem.vue
```diff
@@ -1,10 +1,16 @@
...
+		:srcdoc="html"
+		width="100%"
+		frameborder="0"
+	></iframe>
 </template>
 
 <script lang="ts">
-import type { Ref } from 'vue';
-import { defineComponent, ref, watch } from 'vue';
+import { defineComponent, Ref, ref, watch } from 'vue';
```
drgrice1, line 11: Note that all typescript types should be imported separately from imported javascript objects with `import type` the way I set this before.  The type is then completely erased from the generated javascript code.  This is the case as of version 3.8 of typescript.  In the future eslint will complain about this not being done.

This should be done with all type imports throughout the code.

Although I should also add that rarely do you need to import the vue `Ref` type.  Instead the ref template should be used or for basic types just type inference.  Instead of `const html: Ref<string> = ref('');` this should simply be `const html = ref('');`.  Type inference will do the correct thing here. This is valid for string, boolean, and numeric values.  For more complex types use the ref template.  For example on line 70 of src/components/instructor/ClasslistManager.vue instead of `const selected: Ref<Array<User>> = ref([]);` this should be `const selected = ref<Array<User>>([]);`.  This is all in Vue's documentation.  See https://v3.vuejs.org/guide/typescript-support.html#typing-refs.

drdrew42: Looks like I botched the merge conflict... 🤦

drgrice1: No problem.  I really just wanted to take the opportunity to point out the correct `ref` usage for @pstaabp.

drdrew42: Good for me to see too -- I wasn't aware of this: `const selected = ref<Array<User>>([]);`

pstaabp: I didn't realize about the difference in importing types and js functions/objects.  I'll also move to this.

drdrew42:
## Updates
* pull api request into separate folder in /src
* create constants file
* insert form-submit listener on iframe load

## Unresolved
* overriding request form-data should not happen on the frontend
* two options:
  * pass submissions through the backend (MITM)
  * generate JWE through backend so that requests can be passed directly to renderer

src/components/common/Problem.vue
```diff
@@ -24,49 +37,105 @@ export default defineComponent({
...
+					/* eslint-enable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call */
+				});
+			// .catch();  // need to figure out a more robust way of handling an error like this.
+		}
+
+		async function submitHandler(form: HTMLFormElement, btn: HTMLButtonElement) {
+			const submitUrl = form.getAttribute('action') ?? '/renderer/render-api';
+			// submitAction is a global function from renderer - prepares for submit
+			const submitAction = (window as Window).submitAction;
+       		if(typeof submitAction === 'function') submitAction();
```
drgrice1, line 85: This is what I saw that made me wonder if you knew that the global submitAction of webwork no longer exists. These lines will not work and are also not needed.  When the form is submitted the event listeners attached in ww_applet_support.js will run.

drdrew42:
No longer needed now that #47 is merged

PR #46, opened 2021-09-14
Fix typscript unhandled diagnostic issue

drgrice1:
Fix the typescript unhandled diagnostic that occurs when the quasar dev server is run while the dist folder from the build is present.  This is caused by an incorrect exclusion in the @quasar/app/tsconfig-preset.json file.  In that file the directory is excluded as "/dist".  The forward slash does not work here.  So the exclusion is overridden in our tsconfig.json file.

This is the issue that @pstaabp saw when testing @drdrew42's no-iframe pull request.  I had of course already seen that, but hadn't found the fix yet.  Here it is!

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -14,7 +14,6 @@ module.exports = {
 		'no-invalid-position-at-import-rule': null,
 		'selector-pseudo-class-no-unknown': [true, {
 			ignorePseudoClasses: ['deep', 'v-deep']
-		}],
-
+		}]
 	}
 }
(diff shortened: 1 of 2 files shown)


drgrice1:
To test this first build the quasar project with `quasar build` or `npm run build`, and then run the quasar dev server with `quasar dev` or `npm run serve`.  The dev server should build and then run all of its checks, and report "No issues found".

pstaabp:
Works as indicated.  Going to merge.

PR #49, opened 2021-09-23
Allow trailing commas,

drgrice1:
Although I don't like it, I am removing the eslint rule that enforces no trailing commas as per @drdrew42's request,


drdrew42 (approved):
🥳 ,

PR #50, opened 2021-09-24
Add frontend logging to console + backend log file

drdrew42:
This PR adds the winston logger to the frontend client. 

## Current Configuration
* When not in production, all frontend logs will be routed to the console.
* When in production, no frontend logs will be shown on the console.
* When in production, backend logs will be logged to `logs/webwork3.log`
* Always: logger.error will be routed to the backend and recorded in `logs/clientLog.log`

## Caveats
* new npm modules -- make sure to `npm install`

### Backend
* browser info
* current url
* contents of user's cookie

### Frontend

```javascript
import { logger } from 'src/boot/logger';

logger.info('logger loaded');
logger.warm('dangerous thing');

// also sent to back end
logger.error('really bad thing');
```


src/utilities/logger.ts
```diff
@@ -0,0 +1,38 @@
...
+			port: 8080,
+			path: '/webwork3/api/utility/client-logs',
+			handleExceptions: true
+		})
+	]
+});
+
+//
+// If we're not in production then log to the console
+//
```
drgrice1, line 27: I despise this style of commenting, or any time comments have pre and post comment lines.  Syntax highlighting does a great job of offsetting comments.  There is no need for anything more.

lib/WeBWorK3/Controller/Utility.pm
```diff
@@ -0,0 +1,31 @@
+package WeBWorK3::Controller::Utility;
+use warnings;
+use strict;
+
+use Mojo::Base 'Mojolicious::Controller', -signatures;
+use Mojo::File qw(path);
+use Mojo::Log;
+use JSON qw(decode_json);
+
+my $path = path("$ENV{WW3_ROOT}/logs")->make_path->child('clientLog.log')->touch;
```
drgrice1, line 10: Again, no need to touch.

lib/WeBWorK3.pm
```diff
@@ -134,4 +140,9 @@ sub settingsRoutes($self) {
 	return;
 }
 
+sub utilityRoutes($self) {
+	$self->routes->post('/webwork3/api/utility/client-logs')->to("Utility#clientLog");
```
drgrice1, line 144: I would rather drop `utility` from this route.

drgrice1 (commented):
Good job figuring out how to import winston.  This is a good start for getting logging set up.

A few observations though.

I would rather see the file Utility.pm named something like `Logger.pm`.  I would rather avoid general purpose utility type files.

The `logger.ts` file should be moved from the `src/utilities` directory into `boot`, and also add a logger to the `globalProperties` object as is done in `boot/axios.ts`.

The `winston.format.colorize({ all: true })` option does not seems to work well in the console in all browsers.  Testing in Firefox I get `�[32minfo�[39m: �[32mHello�[39m`.  The ansi escape codes do not seem to be supported in the javascript console.  The colors do work in Chrome.

drgrice1:
One other thing that I noticed, the client logs do not seem to actually get written to the file, and I do not see any thing in the morbo console about the client-logs route being routed to.  Is there something that needs to be done to make that happen?

drgrice1:
Never mind on the logs not being written to the file.  I didn't see that you only have that enabled for `error` logs.  I see that `error` logs do work.

drdrew42:
Oh yeah... I didn't want to bombard the endpoint with anything short of a catastrophe 😂

drgrice1:
Yeah, that is good.

drdrew42:
> I would rather see the file Utility.pm named something like `Logger.pm`.  I would rather avoid general purpose utility type files.

Moved. 

> The `logger.ts` file should be moved from the `src/utilities` directory into `boot`, and also add a logger to the `globalProperties` object as is done in `boot/axios.ts`.

Also moved. 

> The `winston.format.colorize({ all: true })` option does not seems to work well in the console in all browsers.  Testing in Firefox I get `�[32minfo�[39m: �[32mHello�[39m`.  The ansi escape codes do not seem to be supported in the javascript console.  The colors do work in Chrome.

I left this one in. I'll look to see if there's a cross-browser solution. Maybe the winston template accepts metadata as a parameter.

drgrice1:
I put in a PR to this one to do what I meant about the options api instance.

Also, I changed it so that the base url is taken from the VUE_ROUTER_BASE environment variable.  This should be done this way.

drgrice1:
There are a few things that will still need to be figured out.  Clearly the hard coded host and port for the html transport will need to be dealt with.

drgrice1:
I will be sure to use the warm logs most via `logger.warm`.  That should make everyone feel all warm and fuzzy.

pstaabp:
Overall, this looks like a great start.  I haven't successfully run things in production mode yet, but will get to that. 

At some point, we may want to include a log_level parameter which will given the ceiling (or floor) for levels that are output.  

Also, upon reading the documentation, Winston seems to allow logging to files (is this mainly for running node apps?). I can't think of anyway of dumping client-side logging statements to a file without formally making server requests to ship them back.

drgrice1:
> At some point, we may want to include a log_level parameter which will given the ceiling (or floor) for levels that are output.

The `level` parameter of winston is already what you are talking about.  It is the floor for the outputs for a given transport.

> Also, upon reading the documentation, Winston seems to allow logging to files (is this mainly for running node apps?). I can't think of anyway of dumping client-side logging statements to a file without formally making server requests to ship them back.

Logging to a file is for node usage.  It won't work in the browser.  The http transport will need to be used for any logs that we want permanently saved.

pstaabp (approved):
Even though I haven't fully tested the production side of this, I would approve this to be pulled in.

drgrice1:
On the log level thing...

I think that the development log level (in the console) should be set to the lowest priority (or the next to lowest, perhaps we don't need the `silly` level).  The point is that currently the level is set to `info`, so `debug` logs are dropped (as well as `http` and `verbose`).

src/boot/logger.ts
```diff
@@ -0,0 +1,34 @@
...
+import 'setimmediate';
+
+const logger = winston.createLogger({
+	level: 'info',
+	format: winston.format.simple(),
+	defaultMeta: {
+		get location() { return window.location.href; },
+		userAgent: window.navigator.userAgent,
+		service: 'webwork3'
+	},
```
drgrice1, line 11: Is there a way to change this so that this meta data is only logged for the http transport. This is useless information for the developer build in the console, and makes the messages rather inflated.

drdrew42: Yep -- I've simplified the console logs to avoid the default metadata

drgrice1:
Sorry, one more pull request to this.  I forgot to add the logger to the quasar.conf.js boot configuration.

drdrew42:
> Sorry, one more pull request to this. I forgot to add the logger to the quasar.conf.js boot configuration.

That makes sense -- of course simply moving the file isn't enough...

Are we ready to merge this?

drgrice1:
Yeah, I think this can be merged.  The only things that I would like to see resolved are the color issue and removing the defaultMeta data from the console logs.  Those are things that can be resolved later though.

After looking into the color issue more, it seems that in fact Chrome is the only browser that supports this kind of coloring.  Other browsers support using css for color.  Chrome also supports css for this.  So really winston should change their method for browser console logging.  Of course the ansi escape method is used because it was designed for console logs in a terminal, and not in a browser.  Of course css won't work there.

PR #54, opened 2021-09-25
Update the build and deployment method for apache2 on Ubuntu

drgrice1:
This adds an apache2 conf file instead of suggesting that the user copy and paste the configuration into the site configuration file.

Also add systemctl service files to run the webwork3 api and the renderer.

The README.md file is updated with instructions on how to use this setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,9 @@ pnpm-debug.log*
 # Project files
 sample_db.sqlite
 conf/webwork3.yml
+conf/apache2/webwork3-apache2.conf
+conf/apache2/webwork3.service
+conf/apache2/renderer.service
 
 # Devel::Cover
 cover_db/
--- a/README.md
+++ b/README.md
@@ -109,45 +109,62 @@ This section builds all of the UI code using webpack and fires up a browser wind
 4. You should get a "Login to WeBWorK" screen.  You can use the Lisa Simpson account with the username `lisa` and the
    password `lisa`.
 
-### Production build and deployment (instructions for apache2 on Ubuntu2)
+### Production build and deployment (instructions for apache2 on Ubuntu)
 
 TODO: add instructions for other servers and operating systems and add a docker deployment approach
 
 1. Inside the `webwork3` directory, execute `yarn install` or `npm install`.
 
-2. Build the client side user interface with `quasar build`.
+2. Build the client side user interface with `quasar build` or `npm run build`.
 
 3. Copy `webwork3/dist/spa` to `/var/www/html/webwork3` (or create a link).
 
-4. Add the following lines to your apache2 site configuration file:
+4. Copy `conf/apache2/webwork3-apache2.dist.conf` to `conf/apache2/webwork3-apache2.conf`, and create a link to that
+   file in `/etc/apache2/conf-enabled`.  This can be accomplished by executing the following commands from the webwork3
+   directory.
 
-```apacheconf
-# Vue Router configuration for webwork3
-<IfModule mod_rewrite.c>
-    <Directory "/var/www/html/webwork3">
-    RewriteEngine On
-    RewriteBase /webwork3/
-    RewriteRule ^webwork3/index\.html$ - [L]
-    RewriteCond %{REQUEST_FILENAME} !-f
-    RewriteCond %{REQUEST_FILENAME} !-d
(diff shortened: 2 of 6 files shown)


pstaabp:
I have a ubuntu test environment using 20.04 LTS.  The node version is 10.19 and quasar reports it needs node 12.22.1 or greater.  Since 20.04 LTS is pretty standard right now until 22.04 LTS comes out, we should add instructions to upgrade node.  

@drgrice1  Are you running 20.04 or a newer non-LTS version?

drgrice1:
I am running Ubuntu 20.04.  I had to install a newer node version.  I will add instructions later.

pstaabp:
Upgrading node will also handle the conflict between the v1 and v2 lock files.

pstaabp:
I'm also getting an error with the `quasar build` command on ubuntu.  

```
/opt/webwork3/node_modules/fork-ts-checker-webpack-plugin/lib/rpc/rpc-ipc/RpcIpcMessagePort.js:19
        closedError = new RpcIpcMessagePortClosedError_1.RpcIpcMessagePortClosedError(code
                      ^

[  RpcIpcMessagePortClosedError: Process 14504 exited [SIGKILL].
  
  - RpcIpcMessagePort.js:19 
    [webwork3]/[fork-ts-checker-webpack-plugin]/lib/rpc/rpc-ipc/RpcIpcMessagePort.js:19:23
  
  - Generator.next
  
  - RpcIpcMessagePort.js:8 
    [webwork3]/[fork-ts-checker-webpack-plugin]/lib/rpc/rpc-ipc/RpcIpcMessagePort.js:8:71
  
  - new Promise
  
  - RpcIpcMessagePort.js:4 __awaiter
    [webwork3]/[fork-ts-checker-webpack-plugin]/lib/rpc/rpc-ipc/RpcIpcMessagePort.js:4:12
  
  - RpcIpcMessagePort.js:18 ChildProcess.handleExit
    [webwork3]/[fork-ts-checker-webpack-plugin]/lib/rpc/rpc-ipc/RpcIpcMessagePort.js:18:42
  
  - node:events:390 ChildProcess.emit
    node:events:390:28
  
  - child_process:290 Process.ChildProcess._handle.onexit
    node:internal/child_process:290:12
  
] {
  code: null,
  signal: 'SIGKILL'
}
```

I'm guessing you haven't seen this.  I'm using 3.1.0 of quasar and version 16.10.0 of node. Any thoughts?

pstaabp:
A thought might be that this current codebase is quasar 2.0.4 and I just installed quasar on the ubuntu server which is 2.1.0.  Wonder if that matters.

drgrice1:
The package.json file declares any version that is compatible with version 2.0.4.  2.1.0 is compatible.  I am also using version 3.1.0 of @quasar/app and version 2.1.0 of quasar and do not get the error that you are getting.

drgrice1:
I added instructions on how to install node version 16 on Ubuntu to this pull request.

pstaabp:
About the error, above, it appears that Webpack likes a lot of RAM to do it's thing and there are some documented errors using the `fork-ts-checker-webpack-plugin`.  My test server has only 1 GB RAM :( and I think that's the problem. 

I think we'll eventually set things up to be webpacked on a dev machine (or perhaps on GitHub) and then allow a server to just move over the needed files (like in /dist), correct?

drgrice1:
Yeah, eventually we will set up releases. Those can be built on GitHub.

pstaabp:
I got this working on my development machine (MacOS) with minor adjustments.  The one thing I had to do was to make sure that `proxy_module` and `proxy_http_module` was Loaded in the apache config file (`httpd.conf` -- it's different than in ubuntu).  We may want to add a line to the instructions, unless these are on my default in ubuntu.

drgrice1:
Yeah, I was going to add a bit about Apache modules that are needed, but forgot.  They are not on by default.

drgrice1:
I added instructions on how to enable the necessary modules.  You also need the headers and rewrite modules.  Those may be enabled by default, but the command won't fail if they are already enabled.

pstaabp:
Overall, I think this looks good.  I'll have to do some extra work to test on ubuntu.  If you want me to look more into it, I can, if not I think we can merge.

drdrew42:
I haven't had a chance to test much of anything this week -- but don't let me hold things up.

PR #56, opened 2021-09-30
Simple demo to show that multiple problems can be rendered on the same page

drgrice1:
This is not intended to ever be merged.  This is just to demonstrate the rendering capabilities, and show that multiple problems do work with the problem viewer.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -54,6 +54,14 @@ export const instructor_views = [
 		show_user: false
 	},
 	{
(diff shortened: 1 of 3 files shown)


pstaabp:
Nice!!  Looks like you set the `problemPrefix` property to get the ids separate.  I've started a library browser so can integrate this work into the libraryBrowser.

drdrew42:
🤩

PR #57, opened 2021-10-07
redirect log levels to console.warn and console.error

drdrew42:
Forget about colorizing console logs, and swap in the use of console.warn and console.error
We have the added benefit of getting stack traces that way as well...

* logger.error => console.error
* logger.warn => console.warn
* logger.debug => console.warn


drgrice1 (approved):
This looks good to me.  The linter has now succeeded!

drgrice1:
Nope.  Go ahead.

PR #59, opened 2021-10-14
Add perltidy to lint workflow

drgrice1:
The first commit adds perltidy to the workflow.

The second commit just runs perltidy on the code base.

Note that you need the latest version of perltidy from cpan for this (20210717).  The Ubuntu version 20190601-1 does not work the same.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,8 +1,6 @@
 ---
 ###########################
-###########################
 ## Linter GitHub Actions ##
-###########################
 ###########################
 name: Lint Code Base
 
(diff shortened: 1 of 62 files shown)


drgrice1:
Please add any comments/discussion for changes to the .perltidyrc file.  We want this to be set before this is merged.  We definitely don't want this to change much.  Every time that is done the entire code base will need to have perltidy run on, and that will get messy.

drgrice1:
@pstaapb:  Is there any reason why you don't like the `-nbot` rule?

drdrew42 (approved):

PR #63, opened 2021-11-11
Problem set work

pstaabp:
This improves the problem set details view of webwork3 as well as some backend work. Things to consider

- quasar was upgraded, so do `npm install`
- The library browser has the ability to add problems to a set. (Add a few to any problem set)
- The Problem Set Details now has 3 tabs:
   - Details page (setting open/due dates)
   - problems (lists of the problems in the set and allows for reordering)
   - assigning users to a problem set

Note: PR #58 needs to be considered before this.   This seems to be very large, because it is based on #58.


pstaabp:
Pulled in #64 into this branch first and did some merging stuff. 

@drdrew42 I think we should work on the Problem class situation a bit before merging.  After thinking, having a parent class called Problem is the right thing to do, but I think it should have the minimal amount of common fields.  

Another thing to add once the Problem class situation is set is to add some unit testing to this branch. 

I do think it is a good idea to have LibraryProblem and SetProblem inherit from the Problem class, but wondering about HWproblem, QuizProblem, ReviewProblem.  Seems like having SetProblems that are in a HWSet, Quiz, ReviewSet is sufficient and that should be enough to detect the type of problem and render/handling the input/output of such a problem. 

However, if there is something that needs to be set a the problem level (that is not at the ProblemSet level), what would it be?

This being said, I think the following work and needs testing:

- adding a problem from the OPL to a set
- viewing SetDetails, Problems and assigned users from HomeworkSets, Quizzes and ReviewSets
- reordering problems inside a set
- assigning/unassigning users from a set
- overriding dates on a per-user basis.

drdrew42:
I was thinking that `SetProblems` would be used just for manipulation and display as parts of sets (more-or-less restricted to sets detail/sets editor). Student interactions with HW/Quiz/Review problems are expected to differ, both in the sense of rendering options (review: hints/solutions on) and in terms of maintaining state and response submission (quiz/tests are submitted as a batch). But all of these problem objects still must be able to render so they should be able to provide their own RendererRequest params, as well as handle other use-case-specific logic (reordering, in the case of SetProblems).

Right now, the parent Problem class has a lot of fields dedicated to rendering options. These are intended to be set as defaults, rather than being passed in as constructor params. I'm not sure how to handle the dynamic parameter names from `Model` with regards to these fields... I'd rather see them set as an aggregate object, under a single parameter name.

pstaabp:
Just updated again to include deleting problems from a set. Not sure that reordering is fully working, but trying to determine steps to reproduce.

pstaabp:
Reordering should now work.  Reordering in this way requires a new hit to the server for each problem that changed numbers.  I did have a way to do this without a server hit.  If this seems too slow, I'll bring that method back.

drgrice1:
You need to stop upgrading Quasar and the libraries that we are using every time that a new version comes out.  That is not a good idea.  We need to get versions locked in, and only occasionally update.

drgrice1:
I put a pull request in to this branch that fixes the build issues (and deletes a super-linter.log file that was accidentally committed).

pstaabp:
Seems like the `tests-on-separate-db` branch ended up merged into this.  I'll be much more careful with commits going forward.  

@drgrice1 would you like me to pull these that out so we can consider it next?  Or I guess we can consider #61 next and after that pulls in, those changes will be in main and won't be in this branch.

drgrice1:
It would be best to consider #61 next then.  This one is then dependent on that one.

pstaabp:
> It would be best to consider #61 next then. This one is then dependent on that one.

Sounds good.  I think all of the functionality to this branch is back-- a few merge conflicts, but fixed now. 

The Problem class could use a little work, and we need to add unit tests for Problem, LibraryProblem and SetProblem.

drgrice1:
It looks like the main branch merges cleanly into this one.  So that is all you need to do to fix the conflicts.

drgrice1:
I added a pull request to this that removes the eslint-disable statements in the store files, and replaces the any type declarations with more specific type declarations.  Better should be done eventually, but this at least pacifies eslint for now.

pstaabp:
Just saw your PR. Will look soon.  This last commit moves api requests into src/api-requests directory and includes library functionality, session and problem rendering.  Also refactored OPL library browser to not use vuex store.  Definitely this is much cleaner.

pstaabp:
Wish I would have seen your PR earlier.  Reworking the directory introduced merge conflicts, but I think I can take care of these.

drgrice1:
I added another pull request to this one that fixes the jest unit tests. The logic was not quite correct for the setting of defaults in parserParams.

See my comment in that pull request.  We need to discuss some typing things done here.

pstaabp:
I'm not sure what's up.  I'm not seeing this merge conflict at all on my local end, so not sure how to fix it.

drgrice1:
I added another pull request that fixes the merge conflict.  I am not sure why that didn't show up before.  Just an issue with the package.json file and the jquery/bootstrap dependency change.

drgrice1:
Try the pull request that I put in now.  See if this one works.

PR #64, opened 2021-11-12
Refactoring pages + problems

drdrew42:
This is an branch off of the unmerged #58 pstaabp/webwork3:create-model-class
## Changes
- As discussed in our weekly, pages belong in src/pages. I've moved all relevant pages, and updated the quasar tags to be `<q-page>` for consistency.
- I've also added a backend (ww3) route to fetch all problems in a set.
- Using @pstaabp's Model class, I've built out a base Problem class and extended it for LibraryProblem and SetProblem so far.
- LibPanelOPL now uses the LibraryProblem class (through the vuex store)

### Problem Class caveats

Returned library request data has a different structure than what we get from ww3, so Problem objects have a method to return their path (so there is consistency across all Problem object classes). Each Problem also has a method for generating renderer request form-data overrides, with defaults set in the base Problem class and customized in each subclass.

### Problem.vue caveats

I'm currently using `cloneDeep` to maintain a 'free' copy of the problem within the Problem.vue component. We should think about whether or not this is the right strategy long term. I have it that way at present because I don't want to deal with vuex mutators just to re-randomize a problem in the library browser. I also suspect that the sets editor will benefit from de-coupling from the store as we don't want to necessarily pass along each change in real-time to the backend. (Although this means 'taking action' to save changes made...) I'm thinking about re-ordering problems or changing the problem paths, we don't want to be triggering problem updates with every modification.


pstaabp:
I think your use of cloneDeep is the right way to go on problems.  It looks like you solved many of the issue I had with it.  

We need to flesh out this difference between a LibraryProblem and a SetProblem.  Wondering it might be overkill to inherit from a common parent class.  

I’ve been thinking that a SetProblem contain just the fields needed to store in the DB.  (those in db/Result/SetProblem.pm) And it points to a LibraryProblem via the problem_params field (library_id or file_path or …)   

Or perhaps your LibraryProblem class is more of a RenderedProblem which is store all of the information needed to render (with the submit buttons and permission).

drdrew42:
Well, it's slightly easier now that the problem_param is renamed `file_path` ... so now we don't need to use a method to provide the value from different sources.

I do, however, think we might use HWProblem, QuizProblem, ReviewProblem (also descending from the parent Problem class). It also may not be necessary in the long run if we start to implement JWTs, as they will handle the renderer settings for us.

It might seem like overkill here at the outset, but I expect that Problem.vue will explode if we don't design Problem objects to handle some of the logic for themselves.

drdrew42:
These changes have been merged into #63

PR #66, opened 2021-11-17
Add the eslint space-infix-ops rule and general clean up

drgrice1:
Fix the remaining incorrect Ref usages.

Update the remaining import statements to use the provided quasar webpack alias.

Fix my mistake with the jest configuration.  I seem to have mixed the two methods I suggested, instead of only using one.  I installed require-json5, but didn't use it.  I did intend to use that, rather than adding all of the aliases in jest.config.js, and so that is done here.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -46,6 +46,7 @@ module.exports = {
 		__statics: 'readonly',
 		process: 'readonly'
 	},
+
 	rules: baseRules,
 
 	overrides: [{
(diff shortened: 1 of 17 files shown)


drgrice1:
Wait a bit with this.

drgrice1:
Okay, this is ready.

drdrew42 (approved):
Thanks for this @drgrice1 -- just so I'm clear on the changes: neither boot/ nor component/ require leading src/. Are those the only two?

And are we looking to merge this in now, and resolve conflicts with #63?

drgrice1:
There are a few others also.  All of the aliases provided by quasar are listed at https://quasar.dev/quasar-cli/handling-webpack#webpack-aliases.

Yeah, I think this should be merged, and then the conflicts in #63 resolved.

PR #69, opened 2021-11-19
Add a directory structure file to docs for discussion.

drgrice1:
Let me know what you think and if you have any suggestions.  I think this is a good start though.  This of course is just the "src" directory structure.  Later we can add broaden that to the true root of the webwork3 source (to include the perl "lib" and etc).


drgrice1:
We probably don't ever want to actually merge this.  It was really just the quickest way I could think of to make it available to you.  It seems you can't attach files in slack?

pstaabp:
1. Let's add the top-level `api-requests` directory for non-store api work. 
2. Should `css` be inside `assets` ?  I could argue either way except putting it in there would be fewer top-level directories.

drgrice1:
api-requests should be in the common (or lib) location. That is exactly what that is for.

I think that css has to be where it is for quasar.

PR #70, opened 2021-12-09
Add a few more eslint rules.

drgrice1:
Make the npm lint script run on all js, ts, and vue files.  Not just those that are in src.  This fixes things in the build .js files as well
as in the tests directory.

Add ts and vue files to the .editorconfig file.

Eslint does not need to ignore the build .js files.  Just add the /* eslint-env node */ setting.

Add the eslint keyword-spacing, arrow-spacing, and @typescript-eslint/space-infix-ops rules.


pstaabp:
I just pulled this in thinking it was harmless, but now running `npm run lint` I just got the following error:

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.5.0

YOUR TYPESCRIPT VERSION: 4.5.2

Please only submit bug reports when using the officially supported version.

=============
```

However running `tsc -v` I get version 4.4.3.  

@drgrice1 I'm guessing you don't see this.

drgrice1:
I am not seeing this.  If I run `npx tsc --version` I get `Version 4.4.4`.  Both the development and production builds run cleanly for me.  It is unlikely that this is actually caused by this pull request either, as the node modules did not change.  The package-lock.json file changed because something was off with it before.  I suspect you have something not quite right on your machine.  I suspect you have a global typescript installation that is interfering in some way.

PR #75, opened 2022-01-14
Update the internationalization setup.

drgrice1:
Add configuration options to the app i18n object to make it work with the vue composition api.  Also add a fallback locale.

Lazy load the locales from yaml files.  We really don't want to load all languages by default.

Make the language selector show the language names instead of the iso language code.  Note that the displayed language name for the selector is always in its own language regardless of the selected language.  This is how it should be so that users can read the language they want.

Finally, add a few more translations.

### DIFF
--- a/conf/course_defaults.yml
+++ b/conf/course_defaults.yml
@@ -29,28 +29,32 @@
   category: general
   doc: Default language for the course
   doc2: >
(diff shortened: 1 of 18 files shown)


drgrice1:
I fixed conflicts from the last pull request merges, and I decided to go ahead and put this in before it goes stale again.  There is much more to be done here, but a lot of depends on getting the course settings in order first.

drdrew42 (approved):

PR #76, opened 2022-01-14
Add a script to setup the ww3 database for production usage and add support for postgres

drgrice1:
For now the script just creates an empty database.  Eventually it would probably be convenient for the script to also create an admin user.

For postgres things need to be quoted.  Either that or we need to change the name of the table `user`.

### DIFF
--- a/bin/setup_db.pl
+++ b/bin/setup_db.pl
@@ -9,7 +9,8 @@ setup_db.pl - Create and setup the webwork3 database for production usage.
 setup_db.pl [options]
 
  Options:
-   -h|--help    Show full help
+   -h|--help             Show full help
+   -u|--postgres-user    Postgres user (only needed for postgres)
 
 =head1 DESCRIPTION
 
@@ -20,6 +21,10 @@ C<conf/webwork3.yml> before running this script.
 
 Note that this script must be run as root to create the database and user for
 mysql or mariadb.
+
+This script must be run as a postgres user that has sufficient permissions to
+create a postgres database and user.  If the postgres user you are using is not
+"postgres", then pass the correct user in the --postgres-user option.
 
 At this time it is assumed that the database host is localhost, and this script
 does not support alternate hosts or ports.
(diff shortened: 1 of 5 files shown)


pstaabp:
I successfully created a mariadb database with the script.  The Postgres was unsuccessful.  This might partly be my unfamiliarity with Postgres.  A few things though.

1. running on MacOS, the Postgres user is `_postgres`.  At least installing with `brew`.  Perhaps we change the `eq "postgres"` on line 104 to a regexp. 
2. I commented out lines 98-104 and it appears that there are enough privileges to query the database.  I got the database names fine.  However when trying to create the database, I get the error: 
  ```
  DBD::Pg::db do failed: ERROR:  permission denied to create database at bin/setup_db.pl line 114.
  ```

3. I'm able to create a database with `psql postgres` then at the psql prompt, `CREATE DATBASE db_name;` worked fine. I didn't seem to need to login as a different user.  
4. Lastly, we may need to check for valid database names.  I put a hyphen in the name for a mariadb database and just got a generic error.  I'm guessing this wasn't allowed for a database name.

dlglin:
I know that in mariadb when using hyphens in table names you need to put the table name in backticks.  I'm guessing the same is true of DB names.

drgrice1:
The reason that you lack permissions to query the database for 2 would be the same reason as 1.  The `postgres` user is used in the script to obtain the connection.  Why does brew use the `_` in the root postgres username.  That is wrong.  In any case, the best option is probably to add an option to the script to use an alternate root username.

When you run `psql postgres` you are changing users.  So you already have logged in as a different user, and then are able to create the database.

I will look into 4.

pstaabp:
Just added an update to your script that adds a Postgres user.  This works for me on MacOS and hopefully works out of the box for you.

drgrice1:
Your pull request does not work for me.  It fails at the DBI->connect line with
```
ERROR: There was an error communicating with postgres.
DBI connect('dbname=postgres','postgres',...) failed: FATAL:  Peer authentication failed for user "postgres" at ./bin/setup_db.pl line 104.
```

drgrice1:
@pstaapb:  You should be able to run this script with `setup_db.pl -u your_username`, where your_username is, well, your username.  Homebrew sets your user up as a superuser for postgres as I understand it.

drgrice1:
@pstaabp: I changed the script again.  Now you should be able to run `./setup_db.pl` without any arguments.  On Ubuntu, you can run `sudo -u postgres ./setup_db.pl`.  Alternately, you can add yourself as a postgres superuser, and then just run `./setup_db.pl`.  The script just uses the current user for the postgres user.  You actually don't need a postgres superuser.  You only need a postgres user that has the permissions to create databases and other users.  The script also checks that your user has these permissions.

pstaabp (approved):
This works well for me now.  Thanks.  Better than I came up with.

pstaabp:
Going to go ahead and pull it.

PR #77, opened 2022-01-14
Replace lodash usage with the exception of cloneDeep

drgrice1:
In most cases native javascript methods were used to replace the lodash methods.  In a few cases replacement methods were written.  Those methods are in src/common/utils.ts.

The files src/constants.ts and src/common.ts were moved into src/common.  common.ts was renamed to views.ts.  Also src/api-requests was moved into src/common.


drgrice1:
Also note that in some places there was no need to actually use lodash or anything.  There were several cases where copies (clone or the odd pick all usage) and copies were not needed.

pstaabp:
Overall looks good.

I'm trying to just go through all pages as is and have found some bugs that are not necessarily linked to this.  I'll put out a PR that fixes these.  I can either put this into your branch or to webwork3/main to be pulled in after this goes through. 

Also, I'd recommend some documentation on the replacement utilities and perhaps some tests on them.  I'd be happy to do this and give a PR to this branch with that.

drgrice1:
I have noticed several bugs as well.  One is with saving problem set settings.  The ui shows a toast reporting that the settings were changed successfully, but there are errors in the console and reported by the api and the settings are not changed.

If you want to document the replacement utilities and add tests, that would be great.

drgrice1:
There is another obvious issue with the problem set settings.  If you go to change the date or time for a problem set date, a request to the api is sent every single click in the date or time dialog.  That request should only be sent when the date or time is finished being selected. so probably when the dialog closes.  Not for every single click on a date or time in the dialog.

I say dialog here for lack of a better term.  It seems it is a q-menu containing a q-time or q-date, but it looks more like a dialog or popover.  Just UI semantics.

pstaabp:
I have noticed several bugs as well. One is with saving problem set settings. The ui shows a toast reporting that the settings were changed successfully, but there are errors in the console and reported by the api and the settings are not changed.

I see that now. I'll look into this.

> There is another obvious issue with the problem set settings. If you go to change the date or time for a problem set date, a request to the api is sent every single click in the date or time dialog. That request should only be sent when the date or time is finished being selected. so probably when the dialog closes. Not for every single click on a date or time in the dialog.

I've noticed this for a while. I've been trying to determine how to handle this.  I'll think more.

pstaabp:
Doing some testing now.  On `invert`, I'm not getting expected results.  Digging in, it seems like the function:
```
(acc, [key, value]) => ({ ...acc, [value]: key })
```
isn't returning `value` correctly for me.  Is this a newer es format?  I did see you upgraded to es2020, but trying `tsc` on a test is giving me the error:
```
 Type 'string' is not an array type or a string type. Use compiler option '--downlevelIteration' to allow iterating of iterators.

40   Object.keys(obj).reduce((acc: {[key: string]: string}, [key, value]) => {
```

drgrice1:
How are you getting that exactly?  There is nothing used there that needs the newer es.

pstaabp:
I working on a test suite called `utils.ts`.  In there, I basically have a 
```
const obj = { a: '1', b: '2', c: '3', d: '4' };
console.log(invert(obj));
```
which is returning
```
{ undefined: 'd' }
```

If I run `tsc utils.ts`, I get the above error.

> There is nothing used there that needs the newer es.

I hadn't seen `[key, value]` before.   I assume this parses an object in the form `{key: value}` into the key and value.  Though, this is where the error is popping up.

pstaabp:
I don't think this is working in load users from file either.  I thought it might be a jest issue, but don't think so.

drgrice1:
It doesn't seem to be working now.  It was working when I first implemented it, but something has changed with the later changes you made in the problem-set-work branch that was pulled.  I will look at it.

drgrice1:
I see.  Oops.  That should have been Object.entries, not Object.keys.

drgrice1:
Yeah, clearly the `[key, value]` would not work with Object.keys as that parameter would be of type `string`.  It works with Object.entries though as the parameter is an array of type `[string, string]`.

drgrice1:
On a related note the "add users from file" dialog has a problem.  If you open `mlb_list.csv` and then click on a checkbox by one of the users in the list, a "RequiredFieldsException" is thrown with the message "The field(s) 'username' must be present in User", field: "_all", and the ui is left in an unusable state.

If you first click on "First row is header", or just set one of the column headers to "Username", then this does not happen.

pstaabp:
Looks good now.  I should have noticed that should have been `Object.entries` too.

> On a related note the "add users from file" dialog has a problem. If you open mlb_list.csv and then click on a checkbox by one of the users in the list, a "RequiredFieldsException" is thrown with the message "The field(s) 'username' must be present in User", field: "_all", and the ui is left in an unusable state.

I noticed that too.  Working on that in another branch that just has some general bugfixes.

pstaabp:
Looking at `random`, it seems more natural to have `random(min, max)` rather than `random(max,min)`.  This would also be similar to lodash's version.  If okay, I'll change it.

drgrice1:
I went ahead and fixed that here.  That was a mistake.  The usage in problems.ts assumes min is the first parameter, and max the second.

pstaabp:
Just added a PR to this branch for the documentation and testing.

pstaabp (approved):
This looks great.  I think we can rewrite the couple of places in the code that uses `cloneDeep` and then remove lodash.

drdrew42 (approved):

PR #78, opened 2022-01-17
Creation of Student Views

pstaabp:
This makes some templates of the student views.  Currently, this introduces:

- Dashboard (listing of open and closed problem sets--actually shows all problem sets as open)
- Calendar (empty currently)
- Problem Sets (link to all user problem sets)
- Grade (empty currently)

The Problem sets view shows all user set and user problems. Currently only shows the problem, no buttons are present to submit. 

Note: this relies on #73


drdrew42:
So we've now got loads of merge conflicts here -- are these stubs worth saving, or should we recreate them in a new branch?

pstaabp:
I'll take a look at these.  We need to get the update-merge-models branch in before this as well.

pstaabp:
Merge conflicts fixed. And we'll wait until #83 is merged in.

pstaabp:
This is finally in shape to look at.  Mentioned above that it relies on #73, it now relies on #86. 

the functionality is minimal, but a user problem as a student can be viewed.  No submission is working yet. The goal is to just get some basic framework in place.

drdrew42 (approved):
👍

drgrice1 (commented):
There is something wrong with the left nav menu with this pull request.  It is always present.  Even when you first load the site.  It shouldn't be there until after you log in, and should disappear after you log out.  This is a regression introduced in this pull request.

Note that the right menu still stays present when you log out assuming it was visible when you select logout.  That isn't new with this pull request, but needs to be fixed.

There are really a lot of issues with the user interface.  It is unintuitive and really completely disfunctional.  It is probably time that we start working on making something that actually works so that real testing can be done.  Scripted unit tests are only a facade if there is no actual functionality.

Although it is good to change the perl code to use the proper quoting, it would be best if that were done in a separate pull request where that is the only thing done.  It could be quickly merged.  In general this is the approach that you need to start taking with pull requests.  Separate minor concerns (with big edits) into pull requests where that is all that is done.  Put the actual important code changes into a pull request where it those changes can be easily found.  This is one of the main reasons that progress has been so slow with the webwork3 development.

drdrew42:
I've got changes queued up for fixing the sidebars and routing -- but I don't want to introduce more complications in getting caught up. You are right that the UI will need lots of work, but I think that's also something that we deal with once we get back to baseline.

drdrew42:
And when I say 'queued up' I mean, not out for PR yet, just things that I've already worked out (but held back to hopefully make merging easier for now)

drgrice1:
Well, I guess in the interest of getting on with things we can just merge this.  There is a real lack of quality control with the current progression however.

One thing that I note in the pull request is the usage of path parameters in an insecure an unreliable way that we will need to change.  Generally, I don't think that the user_id should ever be a path parameter that is accessed and used via the stash.  The user_id should be stored in the session (it is in fact) and accessed from there.

PR #79, opened 2022-01-17
Fix some issues with problem rendering.

drgrice1:
First, add the problem buttons back so that problems can be tested again.

Remove the pg javascript.  That will now be loaded by PG itself.  Note that to test this you will need to use the
ww3-rendering-work-plus-more-new branch of my renderer fork.  Make sure to update your git submodules also as the lib/PG branch used is different.  PG requests the basic javascript and css needed.  It is assumed that third party javascript and css that is needed is provided (jquery, bootstrap, and mathjax and technically jquery-ui but we don't do that yet or maybe won't).

Make sure that a problems renderer_params are set from the _request_fields.  This is needed to ensure a proper clone via the
`clone` method.

Fix a bug in the parseParams method.  That should not modify the passed in _params directly.  This is called by the constructor, and so by the clone method which calls the constructor for the clone, and that ends up potentically modifying the original problem that was being cloned.


pstaabp:
The attempts table doesn't have any styling.  Should it?  I'm seeing:
image
This is in the Problem Viewer, but looks similar in the Problem Library.

However, overall everything works.  I would only add that in the Library browser, the "Submit Answer" Button should probably be a "Check Answer" button.

drgrice1:
Make sure that you have the correct branch of the renderer.  Also, I forgot to mention that you need to run `npm install` in the renderer/lib/WeBWorK/htdocs directory.  I used the same problem.scss file from my webwork2 bootstrap 5 pull request.  That will eventually be part of pg, and used from the pg/htdocs location.

pstaabp:
That fixes the answer/attempts table.

pstaabp (approved):

PR #80, opened 2022-01-23
Rethinking client-side Models

pstaabp:
This is a rethinking of the client-side models.  Instead of a ModelFactory, this uses a simpler super class (not as a factory) and builds the field and types directly.  The fields of a Model can now be set directly using a setter--and still gets checked to ensure the value is valid or an error thrown.

There is now only a `Model` class with some common functionality like `toObject` and `clone` to be overridden.

Some of the changes:

- `store/models/users.ts` shows how to build up a simple User class.
- `store/models/renderer.ts` contains the Renderer params.
- `store/models/problems.ts` contains a `Problem` class which holds the RenderParams.
as well as a `LibraryProblem` class.
- The Problem viewer in `components/common/ProblemViewer.vue` contains some minimal changes to work with the new `Problem` class.

Overall, please let me know here what you think.  If this looks okay to move forward, the plan will be to:

- [x] Switch all of the other classes in `store/model` to this form.
- [x] Make sure the tests in `tests/unit-tests` still work.
- [x] Try to pull some of the functionality in the `components/common/Problem.vue` file into the Renderer code.  I think this will simplify things overall.
- [ ] add more documentation.
- [x] clean up the differences betweeen the `store/models/renderer-new.ts` and the interfaces in `common/api-requests/renderer.ts`. There are duplicate interfaces.


drgrice1:
I have been thinking about this, and I think one of the things that bothers me (both with the existing models and with the change here) is that there are two types of objects here.  Those are store model objects and api request responses.  The problem is that both of these setups try to put that all into one model, but those are separate entities and can be used separately.  I think that what we need is really separate  classes for handling api request responses, and for store models.  Store models could either derive from the api response classes, or be constructed from them.  Probably the latter, since that will make store model usage more straightforward.  Then other places could also use the api request classes directly in a more convenient way. 

One example of the problem here is with "problems".  We have problems that will have information saved in the store (student problems and set problems), and problems that shouldn't have anything to do with the store (library problems).  Currently this is a big mess as library problems use store model classes.  That doesn't make sense.

pstaabp:
> ...  is that there are two types of objects here. Those are store model objects and api request responses. 

My vision is that all client-side data are stored in some kind of Model whether or not is goes in the store.  Perhaps, we shouldn't have the code for models inside the `store` directory, but instead either a `src/models` directory or somewhere else.

Part of the reason for this rethink is that all problems will be rendered at some point and all the render code (with request params) be put in a common location and that would naturally be in a class.   I think I agree with you that we can have a `Problem` class that handles the rendering and then for example a `LibraryProblem` would basically just render the problem and would not appear in the store, whereas a `UserProblem` would be in the store, but still use the fundamental rendering from a `Problem`. 

Also, as I mentioned in a slack post, I'd like to look at Pinia (https://pinia.vuejs.org) as a vuex store replacement.  Pinia looks quite a bit simpler in that there isn't a complete store for everything--instead each collection of data is its own store and can easier mix and match between data in different stores as well as data in a component that doesn't exist in a store.

drgrice1:
Yeah, that is what I basically mean.  The models are not really in the correct place.  Perhaps they should bin `src/common/api-models` or something like that?

I think Pinia is worth looking at.

pstaabp:
> Yeah, that is what I basically mean. The models are not really in the correct place. Perhaps they should bin src/common/api-models or something like that?

Maybe I'll try to rework this branch and move the new models over to `src/common/api-models` and see what you think.

pstaabp:
Moved the models into `src/common/models` and have ProblemViewer and LibraryBrowser working.  I'm thinking of ditching the ModelParams for any model parameters and just directly make a class with getters/setters.  I thought it would be a lot of work, but it's much nicer. 

The plan will be to now switch everything over to this new style of models, finishing rewriting the tests and test the interface.  

Again, jump in with any comments.

drgrice1:
Sounds good.  Getters/setters is a better way to go.

pstaabp:
This is ready to be tested.  I did the checklist above--although hope to add more documentation.   Overall, it's another huge set of changes, but mainly because all of the underlying client-side model have been completely redone.  I think this results in cleaner and easier-to-understand code. 

I have also gone through all of the pages/views and done some bugfixes and UI reworking.

src/components/instructor/SetDetails/HomeworkSet.vue
```diff
@@ -29,35 +29,42 @@
...
 import { useQuasar } from 'quasar';
-import { cloneDeep } from 'lodash-es';
+import { useRoute } from 'vue-router';
+import { parseRouteSetID } from 'src/router/utils';
 
 import HomeworkDates from './HomeworkDates.vue';
-import { HomeworkSet } from 'src/store/models/problem_sets';
+import { HomeworkSet } from 'src/common/models/problem_sets';
 import { useStore } from 'src/store';
+import { cloneDeep } from 'lodash';
```
drdrew42, line 40: 😢

src/components/instructor/SetDetails/SetUsers.vue
```diff
@@ -118,42 +290,67 @@ export default defineComponent({
...
+				// }
+				// );
+				// } else if (problem_set.value.set_type === 'REVIEW') {
+				// columns.splice(columns.length, 0,
+				// { name: 'open_date', label: 'Open Date', format: formatTheDate,
+				// field: (row: MergedUserReviewSet) => row.set_dates ? (row.set_dates).open : 0 },
+				// { name: 'closed_date', label: 'Closed Date', format: formatTheDate,
+				// field: (row: MergedUserReviewSet) => row.set_dates ? (row.set_dates).closed : 0 },
+				// );
+				// }
```
drdrew42, line 352: Is all this to be removed? Or yet to be updated?

src/pages/instructor/SetDetailContainer.vue
```diff
@@ -55,36 +61,54 @@ export default defineComponent({
...
 		}
-		updateSetID();
 
-		watch(() => route.fullPath, updateSetID);
+		// if (selected_set.value) {
+		// updateSet();
+		// }
+
+		// updateSet();
+
```
drdrew42, line 95: cruft?

src/pages/instructor/SetDetailContainer.vue
```diff
@@ -55,36 +61,54 @@ export default defineComponent({
...
-			const set_id = Array.isArray(s) ? parseInt(s[0]) : parseInt(s);
-			selected_set_id.value = set_id;
+		const problem_sets_info = computed(() => store.state.problem_sets.problem_sets
+			.map(set => ({ label: set.set_name, value: set.set_id })));
+
+		// const updateSet = (_set_id: number) => {
+		// void router.push({ name: 'ProblemSetDetails', params: { set_id: parseInt(`${_set_id}`) } });
+		// set_details_tab.value = 'details'; // reset the tabs to the first one.
+		// };
+
```
drdrew42, line 76: cruft?

drdrew42 (commented):
These are just from a read-through... I'm going to start testing now.

drdrew42:
There are issues with converting problem sets from one type to another. Certainly with regards to the validation of dates -- e.g. a HW set attempting to convert to a Review set encounters issues because of extraneous date fields (WithDates.pm L#38-9) and mismatch between 'due' and 'closed' fields (L#59-60).

drdrew42:
Navigating directly to courses/#/instructor/sets/# fails to load the requested set (showing a blank set details page without the details/problems/users tabs). Selecting the corresponding set in the sidebar does nothing (since the route is already set to that problemset ID), but selecting the corresponding set in the dropdown loads the set and displays the details/problems/users tabs.

drdrew42:
With a fresh run of `build_db.pl`, the review sets (on the set details page) load with the "H" placard (instead of "R"), and on first-click (in the sidebar) a green toast appears with "the problem set 'Review #1' was successfully updated" as soon as the set loads.

drdrew42:
I'm seeing `Uncaught (in promise) Error: [vuex] do not mutate vuex store state outside mutation handlers` errors when attempting to update dates in the Sets Editor

src/store/modules/problem_sets.ts
```diff
@@ -43,13 +43,15 @@ export default {
...
-				commit('UPDATE_PROBLEM_SET', _set);
-			} else {
-				logger.error(`Problem set #${_set.set_id ?? 0} failed to update properly.`);
-			}
+			// The following is not working.  TODO: fix-me
+			// if (JSON.stringify(set) === JSON.stringify(_set)) {
+			// commit('UPDATE_PROBLEM_SET', _set);
+			// } else {
+			// logger.error(`Problem set #${_set.set_id ?? 0} failed to update properly.`);
+			// }
```
drdrew42, line 53: The reason this isn't working is because the store holds ProblemSets with keys that start with `_`, but the response from the API does not prepend with `_`.

src/common/models/courses.ts
```diff
@@ -0,0 +1,180 @@
...
+export interface ParseableCourse {
+	course_id?: number | string;
+	course_name?: string;
+	visible?: string | number | boolean;
+	course_dates? : ParseableCourseDates;
+}
+
+export interface ParseableCourseDates {
+	start?: string;
+	end?: string;
```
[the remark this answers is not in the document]
pstaabp: I was wondering if we want a date in 'YYYY-MM-DD' format or as a unix timestamp, which seems okay, except that will including time as well.

In the bigger picture, I think the whole database needs a serious think before we finalize.  Some of my thoughts (like start and end) have been "gee, we should store something like this" and I want to make sure that everything that goes in has been thoroughly vetted.

src/components/instructor/SetDetails/HomeworkSet.vue
pstaabp: Still working on this.  I thought that calling the clone method of a client-side model would replicate this, but it doesn't seem to.

src/components/instructor/SetDetails/SetUsers.vue
pstaabp: I think I had pulled this out and never put it back in to test.

src/pages/instructor/SetDetailContainer.vue
pstaabp: Good question.  Let me test some more.

src/common/models/courses.ts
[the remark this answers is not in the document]
drgrice1: I think that using unix timestamps is really the only way to go here.  The problem with strings like 'YYYY-MM-DD' is that they do not localize well.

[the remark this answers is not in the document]
drdrew42: UNIX timestamp is ideal, converted to (and from) components on the frontend for display and modification.

src/store/modules/problem_sets.ts
pstaabp: This is a change from the old to new models.  Every time data comes from the server it needs to be parsed and created as a new model. 

Also, this was one place I started testing that the updates were working and there are many places that we need to do this.  Perhaps we can consolidate a lot of code in one location.  

I also intend to develop some unit testing with the server.  That is testing that all of the CRUD commands in the store/modules work.  I think this is a bit tricky with Vuex and maybe the Pinia replacement may be better.

pstaabp:
> There are issues with converting problem sets from one type to another. 

I just haven't built that in yet.  If the set type changes, there are a lot of checks that need to occur.  The dates are one place, but also any of the parameters (that will be stored in the set_params field) may differ between types.  

> Navigating directly to courses/#/instructor/sets/# fails to load the requested set (showing a blank set details page without the details/problems/users tabs). Selecting the corresponding set in the sidebar does nothing (since the route is already set to that problemset ID), but selecting the corresponding set in the dropdown loads the set and displays the details/problems/users tabs.

I had noticed this, but thought I had it fixed.  Will look more.

drdrew42:
## Classlist Manager ##

### Importing users from file ###
- When importing users from a file, there's a noticeable delay when selecting/de-selecting rows.
- Selecting a user who has already been imported is recognized, and a yellow warning to that effect is displayed. 👍 
- It is not clear how to close the import from file modal when an already-imported user row is selected (clicking outside the modal works, but there is little margin for that).

### Default Roles ###
It was not immediately apparent to me that I needed to select a role for the users I selected for import.
- the checkbox must also be selected in addition to filling in the dropdown, which feels redundant and potentially confusing if a role is selected but the checkbox is not.
- with no option selected, users are imported without a role (as expected)
- with 'student' selected, 
  - multiple selected users imported with 'student' role 
  - individual selected users imported with no role
- with 'TA' selected, 
  - multiple selected users imported with 'TA' role 
  - individual selected users imported with no role
- with 'instructor' selected, user is imported without a role (not expected) 
  - multiple selected users imported with 'instructor' role 
  - individual selected users imported with no role

### Editing Users ###
- after attempting to import a user with 'instructor' role, I edited the user to manually assign the instructor role. Doing so added a second copy of the user to the table (state management). Hard-refreshing the page removed the duplicate.
  - I cannot replicate the above, so perhaps the issue is related to the prior attempt to load the user from csv.
- State management when editing users
  - edit a user, 
  - assign a _different_ role, 
  - without deselecting the user - edit again, 
  - the _original_ role is pre-selected, 
  - again assign the _different_ role, 
  - deselect user and reselect user, 
  - edit again - now the _new_ role is preselected.

src/common/models/courses.ts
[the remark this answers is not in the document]
pstaabp: Changed to unix timestamp on next push.

src/common/models/parsers.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+export const mailRE = /^[\w.]+@([a-zA-Z_.]+)+\.[a-zA-Z]{2,9}$/;
+export const usernameRE = /^[_a-zA-Z]([a-zA-Z._0-9])+$/;
+
+export function parseUsername(val: string | undefined) {
+	if (typeof val === 'string' && (mailRE.test(`${val}`) || usernameRE.test(`${val}`))) {
+		return val;
+	} else {
+		throw new UsernameParseException(`The value '${val?.toString() ?? ''}'' is not a value username`);
```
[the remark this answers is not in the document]
pstaabp: good catch.  Fixed on next push.

src/components/instructor/SetDetails/SetUsers.vue
pstaabp: fixed on next push.

src/pages/instructor/SetDetailContainer.vue
pstaabp: cleaned up.

src/pages/instructor/SetDetailContainer.vue
pstaabp: cleaned up.

pstaabp:
@drdrew42 I've fixed a bunch of the bugs you noticed.  I'm going to have to dig in on the performance of the add users by file.  However, if I run that without the javascript console/dev tools open, it is pretty responsive on my laptop. (I've only been using the list of 50 though.)

drdrew42:
## Problem Set Details

### Editing Sets
- problem sets appear with correct placards
- problem set name reports an update on every character entered 
  - we might consider throttling the update
  - name changes are not actually saved
  - name changes are not visible in sidebar or dropdown either
  - refreshing the page reveals updated name
- updating 'open date' to come after other dates displays red `!`
  - but fixing the other dates so that they are valid does not remove the red `!`
  - refreshing the page reveals updated dates without red `!`
- Changing problem set types is not supported (yet)
 
## Classlist Manager

### Adding Users
- load csv, turn on headers, select a user
- red warning appears stating that the role is invalid
  - red warning flag appears on every cell in the row -- somewhat misleading?
- turn on 'import all users as' toggle
- select a role from dropdown -> red warning disappears
- deselect 'import all users as' -> red warning does not reappear
- deselect user and reselect user -> red warning appears
- toggle 'import all users as' back on -> red warning still present despite a selection in the dropdown
- select a _different_ role from dropdown in order to make red warning disappear

### Adding Users Manually
- Attempting to add a user without a role results in an empty red toast at the bottom of the screen - somehow the error message is not getting through
- Attempting to add a user with a negative ID ~~also results in empty error toast~~
- Attempting to reuse an ID ~~also results in empty error toast~~
  - refreshing the page, I get an empty table with "No data available"
  - the user **was** added _without_ a role on the first attempt 
  - subsequent attempts failed because of username uniqueness constraints 
  - the user was added with a negative ID and no user role.
  - missing user role 'UNKNOWN' causes entire table to fail to render (data is sent from backend, but parsers.ts fails)

### Editing Users
- repeatedly editing the same user now displays current status
- editing student ID is not possible (if imported without ID, there is no way to introduce one)

pstaabp:
Just pushed up some new code.  

## Editing Users

- My thought is that instructors can't edit global user information (but we can revisit that).  I added the student ID Information there but not editable. 
- About the student_id.  I have it as a string since IDs could be more than numbers (our students have an @ in them).  Also recitation and section are strings, to be flexible, but we would gain a lot as numbers.  Note: these are more instances that we need to talk details about the DB at some point. 

## Add Users Manually
- There are validation checks for username and role (both required). Perhaps the email needs to be required too unless the username is an email. 

## Adding users from file

- Still thinking of better UI for indicating missing fields (like username/role) and maybe email.

drdrew42:
Closed in favor of #81

PR #81, opened 2022-03-01
Change Vuex store to pinia

pstaabp:
This switches the store from vuex to Pinia, a more lightweight and flexible store option for vue. 

Also, added some functionality to saving course settings, which were never implemented. 

This includes #80 so either this can be pulled in and #80 closed or #80 should be merged first.


drdrew42:
we need to coordinate better... I don't want to sink hours of work just to find out we've got competing solutions. And if your most recent commit gets merged, there's going to be some nasty conflicts to resolve if we want to compare the two.

drdrew42:
@drgrice1 If you're working on reviews, there's a PR to this branch which fixes problem set reactivity -- pstaabp/webwork3#17 (and I'm working on other general cleanup at present).

pstaabp:
we need to coordinate better...

@drdrew42 

> we need to coordinate better... I don't want to sink hours of work just to find out we've got competing solutions. And if your most recent commit gets merged, there's going to be some nasty conflicts to resolve if we want to compare the two.

I agree. After talking I thought it would be a quick fix. 

Generally, when I've started working on something I push a draft PR so everyone can tell what's being worked on.  But in other cases, we can do a quick note in slack.

src/common/api-requests/user.ts
```diff
@@ -0,0 +1,13 @@
+import { api } from 'boot/axios';
+
+import { ParseableCourseUser } from 'src/common/models/users';
+import { ResponseError } from 'src/common/api-requests/interfaces';
+
+export async function checkIfUserExists(course_id: number, username: string) {
+	const url = `courses/${course_id}/users/${username}/exists`;
```
drgrice1, line 7: A general rule of thumb in programming for any language is that if a variable is used only once, it probably should not exist.  Instead just use its value in the one place the variable is used.

You do this a lot.  There are many reasons for this rule.  First is efficiency.  Adding a variable makes the program have to allocate space in the stack for it and then perform a lookup when it is used.  The next reason is readability.  The logic here is actually the same as that for the efficiency reason.  This makes someone reading the code have to look back to find what the variable is.  Yet another reason (seen quite frequently in the code for webwork 2 where this bad programming practice is prevalent) is that as the code evolves that one usage may be removed.  If the coder that removes that usage is not paying attention, the variable definition gets left behind.

I see several places where it seems that you might have done this to prevent long lines.  For example, on lines 30-31 and 36-37 of src/common/api-requests/library.ts.  That is not a valid reason to create a new variable.  Find a way to wrap the line appropriately.  In both of those cases I am easily able to do this.

pstaabp: > I see several places where it seems that you might have done this to prevent long lines. For example, on lines 30-31 and 36-37 of src/common/api-requests/library.ts. That is not a valid reason to create a new variable. Find a way to wrap the line appropriately. In both of those cases I am easily able to do this.

This is exactly the reason I usually have done this.  I can remove the variable and make the line length okay by wrapping.

drgrice1: Sounds good.  Note that one convenience of es6 interpolated strings is that you can always break a line inside of a `${...}`.  For example for lins 36-38 of library.ts you can change those to 
```typescript
	return (await axios.get(`/opl/api/taxo/disciplines/${lib_id.disc_id}/subjects/${
		lib_id.subj_id ?? 0}/chapters`)).data as Array<LibraryCategory>;
```
Note I took this even further and removed the `const response` variable as well.  You don't need to go that far.  Just nothing that this can be taken further.

drgrice1:
It looks like the package-lock.json file did not get properly updated for this pull request.  Running `npm install` changes it.  The file should not have any references to `lodash` and `lodash-es` anymore.

drgrice1:
I am not entirely certain which package change caused this, but with this pull request a newer version of typescript is being installed.  As a result, the versions of the `@typescript-eslint` packages are now complaining that the version of typescript is not supported.  To fix this you will need to upgrade the `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` packages.  The latest versions (5.16.0) work with the newer version of typescript.

pstaabp:
This happened because quasar upgraded when I deleted the package-lock.json file to get rid of lodash.   Working on a different branch I upgraded these packages which gets rid of the warning, but didn't bring that change over here and haven't noticed it since `node_modules` doesn't change between branches.  Just pushed another commit with updated packages.

drgrice1 (changes requested):
I am trying to review this, but the simple fact that nothing works with this pull request makes it really difficult to do so.

There are no problem sets listed on the problem sets manager page (or anywhere for that matter).  There are no users on the classlist manager page.  There is a console log error message about not being able to find the module './.yml'.  I believe that is the localization failing to find the correct yml file because the course language setting is not loading correctly.

Problems do render in the library browser, but we are back to the problem that the mathquill answer boxes are being repeated on the first few problems.

I think that switching to pinia is a good idea, and the getter/setter approach is much nicer than the previous approach with the models, but can we do this without massive regression in what was we had going before?

drdrew42:
I'm not sure what has changed, but there shouldn't be this kind of regression... I specifically added to this PR to get the problem set details page working properly.

drdrew42:
Perhaps a rebuild of your database? I modified the dates for problem sets to use unix rather than date strings... perhaps that's the issue with sets not appearing? I'm not sure about the users page, but mine is loading fine after pulling this branch, `npm install` and rebuilding the db.

drgrice1:
That doesn't help.  That is always the first thing that I try.  I tried it again, and still nothing.  Switching back to main everything works.  This branch ... nothing.

drgrice1:
Morbo is giving the same error repeatedly.  "You must pass exactly one of user_id, username, email."

drgrice1:
@drdrew42:  Which vuex-to-pinia branch are you using for your tests?  If I switch to the vuex-to-pinia branch on your fork, things are a little better.  At least the problem sets are listed.

drdrew42:
I just updated my vuex-to-pinia branch to reflect what I've got working locally. I recall that issue with the morbo backend previously, but I'm pretty sure that's been resolved in this PR.

I'm following pstaabp/vuex-to-pinia, which is the source of this PR.

drgrice1:
Yeah, so I don't know how it works for you.  It certainly doesn't here.  If things are this fragile, then there are problems.

drdrew42:
Before the push, my branch reflected the status of @pstaabp 's PR before pstaabp/webwork3#17. Now it is updated to include my changes, plus resolution of the merge of #73, and the updates to package.json.

drgrice1:
Yeah, now I see that your branch and @pstaabp's are identical.  In any case this pull request isn't working for me.  You say it is working for you?

drdrew42:
Let me try pulling this PR directly... though there shouldn't be any difference? Since your first message reporting this regression, I've rebuilt the db (as you have as well), deleted node_modules and `npm install`ed, so that's the last thing to try, I guess?

drgrice1:
I have tried all of that.  I deleted node_modules, .quasar, and rebuilt the database.  No change.

drdrew42:
After `git fetch upstream pull/81/head:WTF` and `git checkout WTF`, restarting quasar and morbo, this is still working for me. I don't understand why we're seeing different behavior.

drgrice1:
What version of Mojolicious are you running?

drdrew42:
Mojolicious v9.10 -- but that shouldn't be relevant since your errors are coming from `lib/DB/Utils.pm` (and it seems that the request is received with a key mismatch perhaps?)

drgrice1:
It shouldn't matter.  But I will downgrade and see if it makes a difference.

drgrice1:
Okay, well that didn't change anything.

drdrew42:
Running low on ideas here...

drgrice1:
I don't know.  I will try to look at it a little more tomorrow.

lib/WeBWorK3/Hooks.pm
```diff
@@ -43,10 +41,17 @@ sub has_permission ($user, $perm) {
 # Check permission for /api routes
 our $check_permission = sub ($next, $c, $action, $) {
 	return $next->() if ($c->ignore_permissions || $c->req->url->to_string =~ /\/api\/login/x);
+	my $user = $c->current_user;
+	if ($c->stash('course_id')) {
+		my $user_obj    = $c->schema->resultset('User')->getGlobalUser({ user_id => $c->current_user->{user_id} }, 1);
```
drgrice1, line 46: This is the problem.  This should be:
```suggestion
		my $user_obj = $c->schema->resultset('User')
			->getGlobalUser(info => { user_id => $c->current_user->{user_id} }, as_result_set => 1);
```
I don't know how this could possibly be working as it is for you @drdrew42.

drdrew42:
Looks like I still have `ignore_permissions: true` in webwork3.yml... it seems that the underlying config file was updated and I missed pulling in the changes.

drgrice1:
I see.  That makes sense.

drgrice1:
Some logger calls are fine, but some are really not that useful.  In particular the right sidebar messages, and I don't really like the "[MainLayout/updateViews]" messages as those also should be clear.  You can leave them for now, but as development progresses, those should be removed.

Of course the logger messages don't show in production, but even for development if the console is flooded it becomes very difficult to find the things you really need to find.

Okay, so on the alerts I didn't look in the code, and exaggerated.  Some are not native `alert`s, but are native `confirm`s.  Same difference.  Those also need to be replaced.  It seems there are three total.

drdrew42:
library rendering resolved in pstaabp/webwork3#19

pstaabp:
> Okay, so on the alerts I didn't look in the code, and exaggerated. Some are not native alerts, but are native confirms. Same difference. Those also need to be replaced. It seems there are three total.

I was responsible for the confirms and can switch to quasar dialogs.  I guess I did those because they are easy.  Is the reason not to include them a UI preference (i.e. they don't match other quasar dialogs) or something more serious?

drgrice1:
It is something that can wait.  The reason is that native dialogs give a bad user experience because they are ugly and they don't behave well like a quasar dialog.  With a quasar dialog you can click outside the dialog to close it (unless you set the option to not allow that).   I don't think that native dialogs are friendly for assistive technologies either.

src/common/models/renderer.ts
```diff
@@ -0,0 +1,100 @@
...
+import { parseBoolean, parseNonNegInt } from './parsers';
+export interface LibraryParams {
+	library_id?: number;
+	file_path?: string;
+	problem_pool_id?: number;
+}
+
+export interface ParseableRenderParams {
+	problemSeed?: number | string;
+	permission_level?: number | string;
```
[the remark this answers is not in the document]
pstaabp: Fixed in latest commit.

pstaabp:
Finally digging into this.  I have been thinking about the handling of permission on mojolicious which is what the code in the `Hooks.pm` code will eventually handle. However, since it was getting in the way of building out things, I added that `ignore_permissions` in the config file which was skipping that block that we ran into errors.  

Clearly, we will need to revisit the handling of permissions and tighten up all the security.

pstaabp:
Just pushed a few more changes.  Some bugfixes and switched the native alerts/confirms to quasar versions.

drgrice1:
@pstaabp:  Pull request pstaabp/webwork3#20 fixes another regression in this pull request with the language selector in the user settings.

drgrice1:
Once you merge my pull request, then the main regression that is left is to get problem set problems to appear again.  Currently when you open "HW #1" in the "Problem Sets Manager" and then click on "Problems" it says there are no problems in the set.

src/pages/instructor/Instructor.vue
```diff
@@ -5,39 +5,64 @@
...
+
+		const course_id = parseRouteCourseID(route);
+
+		logger.debug('[Intructor]: fetching users from the server.');
+		void users.fetchMergedUsers(course_id);
+
+		logger.debug('[Instructor]: fetch problem_sets from server');
+		void problem_sets.fetchProblemSets(course_id);
+
+		void problem_sets.fetchSetProblems(course_id);
```
drgrice1, line 52: I guess this will do for now, but this really shouldn't be done here.  The problems shouldn't be fetched until a set is actually loaded, and at that time only the problems for that set should be loaded.  Not all problems for all sets.  In fact, the course users and problem sets shouldn't even be loaded at this point.  None of that is needed at this point.  This all needs to be rethought.

On another note, this revealed another regression.  The answer prefix is clearly not being set (or not set uniquely) for the problems when they are loaded for problem sets as the same thing is happening there as was happening in the library browser.  Mathquill answer blanks are all begin created in the first problem.

drdrew42: Agreed... more granularity WRT loading data.

I'll have a look and see about the problem prefixes with @pstaabp's latest commit.

pstaabp: I was thinking that we load data at the beginning if it is needed for multiple views.  

For example, the reason for loading the set problems is that clearly they are needed for set details, but also, I was thinking that the Problem Sets Manager (editor) should show the number of problems in a given set and therefore would be needed there.   There's work to be done on the Problem Sets Manager.

Another approach would be to wait until the data is needed and then load it (either in the store or just locally).   I feel like that could be tricky to make sure that it does get loaded.

drgrice1:
However, that is not the correct way to do it.  What should happen is that data is loaded as needed.  If the store is set up right, then when data is loaded it is added to the store.  If data is requested that is already in the store, then that data is used.  Otherwise, the needed data is fetched.

This will be a massive mistake if left as it is.  Just think about it a little bit.  Say you have a thousand users in a course (perhaps extreme, but not unlikely in many situations), and say you have 100 sets each with 20 problems.  That is a really large amount of data to load all at once here, and just shouldn't be done.

It may be "tricky" (shouldn't really be if the store is worth anything), but it is how it should be done.

pstaabp:
I realize this scale problem and don't load user sets or user problems until needed.  Those are much more local that general problem sets.  

I'm thinking the fetchXXX actions in each store could check if XXX has already been loaded then skip.  If not, then load XXX.  Then calling fetchXXX multiple times will only result in loading data once.  However, we may need a flag to force reload as well.

drdrew42:
Agreed -- streamlining this is something that we need to handle with care, and is not an immediate need.

src/pages/instructor/Instructor.vue
drdrew42: > The answer prefix is clearly not being set (or not set uniquely) for the problems when they are loaded for problem sets

I'm not even getting the problems to fetch from the server... it doesn't look like there's any implementation of `fetchSetProblems` anywhere in the code?

drdrew42: Somehow I ended up missing this line during a merge... `fetchSetProblems` -- fixed now.

drdrew42: opened a PR on pstaab/webwork3#22 to fix mathquill on set details page

PR #82, opened 2022-03-14
Updating Add Course

drdrew42:
This PR should not be considered until #73 and #81 have been merged.

### Resolved
- Modification of course dates is now functional (consistent with existing data structure, but see below)
- Instructor 'user' now resets after changing username from existing user to non-existent user
- `addCourse` updated to send `course.toObject`, resolving API rejection
- UserCourses page now parses `user.role` accurately when filtering (see below)
- refactored `addCourse` modal function to use promise.then.catch

### Unresolved
- consider updating all scripts to ensure consistent role capitalization in DB?
- ~~formatting for `course.course_dates` -- why is it `YYYY-MM-DD` when other DB dates are unix?~~


pstaabp:
@drdrew42 Looks like there are some conflicts now on this branch.

drdrew42:
@pstaabp the conflicts now present in this branch should be resolved once #81 is merged.

pstaabp:
@drdrew42 Just sent a PR to your branch that fixes the failing test for this.

src/components/admin/AddCourse.vue
```diff
@@ -80,58 +77,137 @@ export default defineComponent({
...
+		};
+
+		const addCourse = async () => {
+			let course_instructor: User | undefined;
+
+			// add global user for instructor if they don't exist yet
+			if (!instructor_exists.value) {
+				// unclear why user.value must be re-cast...
+				course_instructor = await users.addUser(user.value as User)
+					.then((_new_user) => {
```
pstaabp, line 115: Shouldn't this be either await or using .then, but not both.  I'm trying to track an error down related to the course dates, which aren't updating.

drdrew42: Promises can be chained regardless of await/void. In this case, we are waiting for the return value of a User ... all that's happening with `.then` is we are peeking into the returned User for the purpose of notifying through the UI.

Specifically, since we are assigning the result of this promise to a variable, we must make sure that our promise chain returns the user if the request was successful - and undef otherwise.

pstaabp:
@drdrew42 I was getting a propType error on the InputWithBlur component and just put in a minor fix to that. 

I think if you're fine with the functionality with this, we can merge everything in.

PR #83, opened 2022-04-01
Add tests for each of the stores and refactor merged users and merged user sets.

pstaabp:
This pull request does a few things
1. Adds tests for each of the stores. 
2. Because "merged" models are difficult to keep in sync with their components, merged models are getters from the store. Refactor all of them. 
3. Some cleanup to the stores for consistency.
4. Refactor the code in the `*.vue` files.


pstaabp:
This is ready to review now.  A couple of additional things
1. The `npm test` script now only runs the tests in the `tests/unit-tests` directory. 
2. The `npm test-stores` script runs the tests in the `tests/stores` directory in a non-parallel fashion.  In parallel, the access to the database is in parallel which messes up the results. 
3. The `npm test-stores` is not run in the GitHub action.  This is because we'd need to set up the server to be ready to handle db api requests.  I realize this is possible, just didn't have time to do this.
4. For some of the vue files, I switched over to [`script setup` format](https://vuejs.org/api/sfc-script-setup.html).  When using the composition API in vue, it is not the recommended way, and is quite nice--simplifying some of the code. 

To test, 
1. run `npm test` to test the models using the unit tests.
2. run `npm test-store` to test the stores.  Make sure that the webwork3 mojolicious (morbo) server is running for this. 
3. run the perl tests `prove -r t`
4. test the functionality of the webwork3 front end.

src/common/models/problem_sets.ts
```diff
@@ -116,37 +118,46 @@ export interface ParseableQuizDates {
 }
 
 export class QuizDates extends Model {
-	private _open = 0;
-	private _due = 0;
-	private _answer = 0;
+	private _open?: number;
+	private _due?: number;
+	private _answer?: number;
```
drdrew42, line 123: I don't think we want to allow any of these to be undef, no? I can see for a UserQuiz that the overrides might be undef, but it's problematic to allow this... 

Plus, this change will now require an undef check anytime we want to access these dates on the frontend. What do we gain in exchange?

src/common/models/problem_sets.ts
```diff
@@ -116,37 +118,46 @@ export interface ParseableQuizDates {
...
+	public set answer(value: number | string | undefined) {
+		if (value != undefined) this._answer = parseNonNegInt(value);
+	}
 
 	public isValid() {
-		return this.open <= this.due && this.due <= this.answer;
+		let valid = true;
+		if (this.open && this.due && this.open > this. due) valid = false;
+		if (this.due && this.answer && this.due > this.answer) valid = false;
+		return valid;
```
drdrew42, line 160:
```suggestion
		if (this.open && this.due && this.open > this. due) return false;
		if (this.due && this.answer && this.due > this.answer) return false;
		return true;
```

src/common/models/problem_sets.ts
```diff
@@ -261,44 +272,54 @@ export interface ParseableHomeworkSetDates {
...
 
 export class HomeworkSetDates extends Model {
-	private _open = 0;
-	private _due = 0;
-	private _reduced_scoring = 0;
-	private _answer = 0;
+	private _open?: number;
+	private _due?: number;
+	private _reduced_scoring?: number;
+	private _answer?: number;
```
drdrew42, line 278: Same comment as above.  Maybe with an exception for reduced_scoring...

src/common/models/problem_sets.ts
```diff
@@ -378,35 +399,41 @@ export interface ParseableReviewSetDates {
 }
 
 export class ReviewSetDates extends Model {
-	private _open = 0;
-	private _closed = 0;
+	private _open?: number;
+	private _closed?: number;
```
drdrew42, line 403: Same.

src/common/models/problem_sets.ts
```diff
@@ -378,35 +399,41 @@ export interface ParseableReviewSetDates {
...
+	public set closed(value: number | string | undefined) {
+		if (value != undefined) this._closed = parseNonNegInt(value);
+	}
 
-	public isValid() {
-		return this.open <= this.closed;
+	public isValid(): boolean {
+		let valid = true;
+		if (this.open && this.closed && this.open > this.closed) valid = false;
+		return valid;
```
drdrew42, line 433:
```suggestion
		if (this.open && this.closed && this.open > this.closed) return false;
		return true;
```

src/common/models/problems.ts
```diff
@@ -94,14 +94,20 @@ class ProblemLocationParams extends Model {
 
 	}
 
-	public get library_id() : number { return this._library_id; }
-	public set library_id(val: string | number) { this._library_id = parseNonNegInt(val);}
+	public get library_id() : number | undefined { return this._library_id; }
+	public set library_id(val: string | number | undefined)
+	{
+		if (val != undefined) this._library_id = parseNonNegInt(val);
+	}
```
drdrew42, line 101: I do not like functions starting with an open brace on a new line. Personal preference?

src/common/models/problems.ts
```diff
@@ -94,14 +94,20 @@ class ProblemLocationParams extends Model {
...
+	public get file_path() : string | undefined { return this._file_path;}
+	public set file_path(value: string | undefined) { this._file_path = value;}
 
-	public get problem_pool_id() : number { return this._problem_pool_id; }
-	public set problem_pool_id(val: string | number) { this._problem_pool_id = parseNonNegInt(val);}
+	public get problem_pool_id() : number | undefined { return this._problem_pool_id; }
+	public set problem_pool_id(val: string | number | undefined)
+	{
+		if (val != undefined) this._problem_pool_id = parseNonNegInt(val);
+	}
```
drdrew42, line 110: personal preference?

src/common/models/user_sets.ts
```diff
@@ -2,34 +2,39 @@
  * This defines UserSets and MergedUserSets
  */
 
-import { Model } from '.';
+import { Dictionary } from 'express-serve-static-core';
+import { generic, Model } from '.';
```
drdrew42, line 6:
```suggestion
import { Model } from '.';
```
`.toObject` later will receive proper typing automatically.

src/common/models/user_sets.ts
```diff
@@ -411,3 +461,56 @@ export function parseMergedUserSet(user_set: ParseableMergedUserSet) {
...
+	const params = set.set_params.toObject() as Dictionary<generic>;
+	const user_params = user_set.set_params.toObject() as Dictionary<generic>;
+	Object.keys(user_params).forEach(key => {
+		if (user_params[key] !== undefined) params[key] = user_params[key];
+	});
+	merged_user_set.set_params = params;
+
+	// override set dates
+	const dates = set.set_dates.toObject() as Dictionary<generic>;
+	const user_dates = user_set.set_dates.toObject() as Dictionary<generic>;
```
drdrew42, line 510:
```suggestion
	const params = set.set_params.toObject();
	const user_params = user_set.set_params.toObject();
	Object.keys(user_params).forEach(key => {
		if (user_params[key] !== undefined) params[key] = user_params[key];
	});
	merged_user_set.set_params = params;

	// override set dates
	const dates = set.set_dates.toObject();
	const user_dates = user_set.set_dates.toObject();
```

drdrew42 (commented):
Incomplete review. Still need to look at instructor components + pages.
Tests run as expected.
Unexpected hiccup: install Volar plugin for VScode and disable Veture.

src/common/models/problem_sets.ts
pstaabp: > I don't think we want to allow any of these to be undef, no? I can see for a UserQuiz that the overrides might be undef, but it's problematic to allow this...

I had some rationale for this when I switched them, but can't remember that right now (I should have added a comment). I think you're right though about having default values for the non-user versions of problem sets.  Will switch back.

src/common/models/problems.ts
pstaabp: I don't why I did this.  I don't think the linter picks up on these.

src/common/models/user_sets.ts
pstaabp: A vscode auto-filling imports sometimes picks up the wrong one.

pstaabp:
> Unexpected hiccup: install Volar plugin for VScode and disable Veture.

I've been use volar lately, since the Vue project recommends this now.  I'm not sure I have the settings correct because it misses some of the type errors.  I'm usually going off of the results of `npm run lint` instead.

drdrew42:
When using Veture, my VScode choked on the refactored `script setup` vue files in particular. `quasar dev` ran the code just fine, but the code editor did NOT like it until I switched to Volar.

src/common/models/problem_sets.ts
pstaabp: I got a little time to think about this.  I had realized that a UserQuiz and Quiz had the same date structure except that the UserQuiz could have undefined values, so that's why I switched.  I have figured out a way to do this in a better way now, so QuizDates cannot have undefined values, a UserQuizDates can have undefined values (and then a MergedQuizDates will also not have undefined values). 

I just rewrote the code this morning and will get to testing now.

src/common/models/problem_sets.ts
```diff
@@ -67,11 +69,11 @@ export class ProblemSet extends Model {
 	}
 
 	set(params: ParseableProblemSet) {
-		if (params.set_id != undefined) this.set_id = params.set_id;
-		if (params.set_visible != undefined) this.set_visible = params.set_visible;
+		if (params.set_id) this.set_id = params.set_id;
+		if (params.set_visible !== undefined) this.set_visible = params.set_visible;
 		if (params.course_id != undefined) this.course_id = params.course_id;
```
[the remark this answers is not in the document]
pstaabp: I've been trying to use truthy checks for string types to prevent setting fields to the empty string '' and otherwise checking != undefined.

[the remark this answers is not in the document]
drdrew42: So, for booleans, we clearly cannot rely on truthy/falsey; and same for any numerical values that _could_ be zero... but beyond that, a truthy check should suffice.

Especially because (type-wise) we are very broad about the types we allow, these undefined checks would permit empty strings...

src/stores/courses.ts
```diff
@@ -9,20 +9,41 @@ export interface CourseState {
...
 		courses: []
 	}),
 	getters: {
-		getCourseByName: (state) => {
-			return (courseName: string) => state.courses.find((course) => course.course_name === courseName);
+		/**
+		 * returns a course in the store.
+		 * @param {CourseInfo} course_info - either a course_id or course_name
+		 * @returns {Course} the requested course
+		 */
```
drgrice1, line 25: This is invalid.  This function takes the state parameter and returns a function that has the course_info parameter of type CourseInfo and returns a Course object.  It is not a function that takes a course_info parameter of type CourseInfo and returns a Course object.

drgrice1 (changes requested):
There are some things that need to be fixed before this is ready.

First, the new store tests do not work.  If you are going to add new tests, they should at least succeed.

Next, there are numerous places where the jsDoc is incorrect and needs to be fixed.  I comment on one of them that was added in this pull request.  There are errors in several other files as well.  Some of those are from earlier pull requests that I apparently didn't review as carefully.

src/stores/courses.ts
pstaabp: I agree that this returns a function, however as a getter, it is called like:
```
course_store.findCourse({course_id: 3 })
```
for example, which is why I wrote the documentation that way.

pstaabp:
> First, the new store tests do not work.

They work for me. 
1. rebuild the test database with `perl t/db/build_db.pl`
2. make sure that the mojolicious server is running for the ww3 database.
3. `npm run test-stores`

src/stores/courses.ts
drgrice1: However, it is invalid for jsDoc.  It has to actually be correct for what it actually is.

drgrice1:
> > First, the new store tests do not work.
> 
> They work for me.
> 
>     1. rebuild the test database with `perl t/db/build_db.pl`
> 
>     2. make sure that the mojolicious server is running for the ww3 database.
> 
>     3. `npm run test-stores`

Nope.  Doesn't work for me.  I am really not sure how it does for you.

drgrice1:
I am willing to bet that you still have ignore_permissions still set in your conf/webwork3.yml file.  YOU NEED TO REMOVE THAT.

Even with that set in my conf file there is still one test suite that fails, and 2 tests that fail.

drgrice1:
Speaking of permissions, the unit-tests that check permissions all fail and need to be fixed as well.

pstaabp:
Yes, I have ignore_permissions set to true in conf/webwork3.yml and honestly, I haven't tested any permissions in many months and haven't tested that either.   I've been trying to get the functionality working without worrying about the permissions right now.  I know eventually I need to get back to that, but can look into that if you'd like that in this PR.

It's odd you're still getting errors with ignore_permissions. Which tests are failing?

drgrice1:
I am going to be removing the ignore_permissions option entirely from the code.  That was a bad idea from the start.  The tests should be using permissions, not ignoring them.

The tests that are failing are when ignore_permissions is 1:
In the user_sets.spec.ts test suite, 'Fetching UserSets > Fetching Use sets from a course' and 'Test merged user sets > Test all merged user sets'.

pstaabp:
My thought was that we hadn't thought out the permissions yet and this was a way to get around that.  Once we had nailed down the permission structure I was planning to get rid of the ignore_permissions.  I think if we ready to dive into now, I'm fine with that. 

All of the test code is basically for instructor level permissions, however I think we have thought about that more fine-grained, so maybe a good plan for all of the tests that use the mojolicious server currently is to log in as an instructor and run all of the tests.  

I haven't figured out how to do this on the client-side yet using jest, so we'll need to build that in as well.

drgrice1:
Also, note that it is a very simple thing to add authentication to the store tests so that it will work with my pull request #90.

pstaabp:
I'll pull in #90.

The time zone.  Yes.  Totally a hack.  I'll update to UTC format and retest.

drgrice1:
Yes, I mean the test in `tests/stores`.  Those do not directly use the yml configuration files, but they require that the webwork3 api is running via morbo which does use the webwork3.yml file.  So the store tests do use them.

pstaabp:
I just made a test branch to pull #90 into this branch and the `test/stores` tests are not working.  I'm also getting some errors on the quasar app with permission errors.  

I think with the `tests/stores` and authentication, we need to store the session in the jest tests. I'm trying to figure this out right now.

drgrice1:
You will need to set `cookie_secure: 0` in webwork3.yml.

Then add `await api.post('login', { username: 'admin', password: 'admin' });` in the beforeAll handler in each test file.  Of course add `import { api } from 'boot/axios';` is needed in each file also.

pstaabp:
This works.  I actually used: 
```
const session_info = await checkPassword({
			username: 'admin',
			password: 'admin'
		});
```
instead.

drgrice1:
Using `checkPassword` from the src/common/models/courses.ts is overkill.  It is best to use what I suggested.  Use the correct tools to do the job.  Don't use a hammer to push in a thumbtack.

drgrice1:
There is no need for the axios cookiejar.  The cookies are set.  Mojolicious takes care of that with the session.  The only difference between the ui in the browser and the tests is that most browsers allow secure cookies for localhost even though it is not secure, and the the tests are more proper.

pstaabp:
I think this is ready to test again.

pstaabp:
I find that the jsDoc style comments are nice in that the editor I use (vscode) dynamically pulls them in as a popups.  

> it can have side effects when used with typescript

What are the side effects?

drgrice1:
Typescript can take the types from the jsDoc and use them.  Generally, the point of jsDoc in relation to typescript is for pure javascript files that are imported into typescript files.  Then typescript uses the types declared there to help in inferring valid types for that file.  However, it can do the same thing when types are given in jsDoc inside a ts file, and that can cause issues.  So it is actually recommended that if you use jsDoc in a ts file, that you do not add the types.

drgrice1:
As to the VSCode popups, if you use a better editor you get those popups inferred by typescript anyway.  Crappy VSCode.

src/stores/courses.ts
drgrice1: If jsDoc is to be used, then this needs to be fixed.  This fails to generate valid documentation here.  Instead the tsserver complains about a @param tag with name course_info and no parameter with that name.  The @return type should be removed also.  That could be a real problem as typescript might try to use that.

This also occurs with some of the jsdoc in set_problems.ts and users.ts.

pstaabp:
VSCode does give info based on typescript, but often that isn't clarifying. And sometimes it gives a model (something that extends Model) a list of all fields/methods, which is often not helpful and the jsdoc code can give useful documentation. 

> So it is actually recommended that if you use jsDoc in a ts file, that you do not add the types.

If this is the case, maybe including short description of classes/functions without the parameters/arguments and return types.

drgrice1:
For now, just remove the @param's and @return's on the ones that are incorrect, and change those to a short description.  Then we can merge this pull request.

pstaabp:
Will clean up the jsdoc stuff then. 

Also, looking into how to start morbo with npm run test.  The https://www.npmjs.com/package/@mojolicious/server-starter doesn't seem to be happy in a typescript environment. Even adding a `.d.ts` file for this, there's something incompatible with typescript.  

I will probably look into a script that 1)  directly starting morbo in a test environment (which we can load the `ww3-dev.yaml` file), 2) running `npm run test-stores` then 3) shutting down morbo.  I'll add this as a separate PR branched off `main` to try to avoid the pyramid of PRs that I have built up.

drgrice1:
Yeah, I looked closer at using the @mojoliciuos/server-starter also, and quickly found that it won't easily work with typescript.

pstaabp:
I think this is all cleaned up.

drgrice1:
Thanks for making the requested changes.  I will merge this now.

PR #86, opened 2022-05-03
Rename models

pstaabp:
This renames a few model names as well as changes the way models are handled in the store.  This includes:
- any model with "Merged" in it, drops the "Merged" in the name MergedCourseUser -> CourseUser
- for models with merged in them, change the "basic" model to prefixed with a "DB".  UserSet->DBUserSet

The idea will be all models outside the stores will use the old "Merged" models, now renamed.  It is much easier to handle the client-side data with these models.  

In the stores, when data needs to be saved (added/updated), send only the DB version of the data. For example, a `UserProblem` now contains information about the user (like username) and the problem set it is in (like the set_name).  This information is not in the database version of the User Problem and only that data is sent. 

This is dependent on #83 and that should be merged first.


pstaabp:
This is ready for review.  It contains all the changes from #83 , so can be reviewed/merged in lieu of that one.

In short, 
- the models (in `src/common/models`) should be examined, especially the CourseUser (now a merge with a global user), all of the UserSets (merge between DBuserSet, a ProblemSet and a CourseUser) and the UserProblem (a merge between a DBUserProblem, SetProblem, UserSet).  
- all tests should run clean (`npm run test`, `npm run test-stores`, `prove -r t`)
- the app/UI should function as before. 

No new functionality was introduced.

pstaabp:
Fixed all merge conflicts since #83 was merged in and some code cleanup. 

Sorry about all of the changes.  I've been working on cleaning up the code and decided to including switching strings without interpolation to 'these types' (not "these types").  There isn't any functionality change on the perl side of things.

The main functionality change in this PR is in `common/models/` then `problems.ts`, `problem_sets.ts`, `user_sets.ts` and `users.ts` to do the name switch.  The UI side then reflects these changes (no new UI functionality).

drdrew42 (approved):
I think this is ready now!

lib/DB/Schema/ResultSet/User.pm
```diff
@@ -392,10 +392,10 @@ sub addCourseUser ($self, %args) {
...
 	my $course_user = $self->getCourseUser(info => $args{info}, as_result_set => 1, skip_throw => 1);
 	DB::Exception::UserAlreadyInCourse->throw(
-		message => "The user with " . $args{info}->{username} ? ("username: " . $args{info}->{username})
-		: ("user_id: " . $args{info}->{user_id}) . " is already in the course with " . $args{params}->{course_name}
-		? ("course name: " . $args{info}->{course_name})
-		: ("course_id: " . $args{info}->{course_id}))
+		message => "The user with $args{info}->{username}" ? "username: $args{info}->{username}"
+		: "user_id: $args{info}->{user_id} is already in the course with $args{params}->{course_name}"
+		? "course name: $args{info}->{course_name}"
+		: "course_id: $args{info}->{course_id}")
```
drgrice1, line 398: @pstaabp:  This is pretty messed up, and needs to be fixed.  The message will always be `"username $args{info}->{username}"` regardless of if `$args{info}->{username}` is defined.

PR #87, opened 2022-05-15
Change problem model name for Problem

pstaabp:
When using the name Problem for the database, we really mean a SetProblem and so we change the name for this on the backend and where it affects the front end.

This depends on #86


pstaabp:
Update: made sure that the addSetProblem method in the set_problem store only adds a LibraryProblem. 
Also, added a test after each of the db tests to ensure that the database state is restored. (& caught a couple of errors.)

Ready for testing again.

drdrew42 (approved):
test run

PR #88, opened 2022-05-16
Changes the server code to send json booleans instead of perl 0/1s

pstaabp:
This is a refactor of database code to ensure that the mojolicious server sends JSON true/false values for all booleans instead of perl 0/1s. This includes fields in each database table as well as in the `params` field which is stored as a JSON.  

To test, run all tests: `prove -r t`, `npm run test` and `npm run test-stores`.

No UI changes are made, so there should be any need to test the UI for changes. 

This depends on #86, so that need to be merged first.

### DIFF
--- a/src/components/instructor/ClasslistManagerComponents/AddUsersFromFile.vue
+++ b/src/components/instructor/ClasslistManagerComponents/AddUsersFromFile.vue
@@ -333,7 +333,7 @@ const addMergedUsers = async () => {
 		}
 		try {
 			await users.addCourseUser(new CourseUser(user));
-			const full_name = `${user.first_name as string } ${user.last_name as string}`;
+			const full_name = `${user.first_name as string} ${user.last_name as string}`;
 			$q.notify({
 				message: `The user ${full_name} was successfully added to the course.`,
 				color: 'green'


lib/DB/Schema/Result/CourseUser.pm
```diff
@@ -88,19 +88,6 @@ __PACKAGE__->table('course_user');
...
-	useMathView    => q{[01]},
-	displayMode    => q{\w+},
-	status         => q{\w},
-	lis_source_did => q{\w*},
-	useWirisEditor => q{[01]},
-	showOldAnswers => q{[01]}
-};
-
-our $REQUIRED_PARAMS = {};
-
```
pstaabp, line 103: This is deleted because it was repeated below as switched to being returned from a subroutine instead of just a hash.

drgrice1:
One downside to this approach is that it makes it so that the JSON database columns are not as flexible.  The idea behind having those columns was so that data could be more freely added to them.

Actually, thinking about this more this isn't good.  The values in the JSON columns are not being entered correctly into the database.  The InflateColumn::Serializer::JSON package already properly deals with boolean values, but the code is not saving them as boolean values.  So instead of the override package that you hacked in, you need to figure out where the JSON is getting the integer values from.  I have tested this, and I see that if values are properly stored in a JSON column as booleans then a dump of the data shows `bless( do{\(my $o = 1)}, 'JSON::PP::Boolean' )`.  That will be properly encoded to 'true' in JSON.

This needs more though.

pstaabp:
With the current code, boolean is the params fields are being saved in the DB as JSON true/false.  I guess you're advocating for keeping them as 0/1 instead.  Is this what you're saying?  If so, why?

The reason for making another Serializer class like this was that it could be handled in a single place.  Otherwise, we were needing to do the conversion in every DBIx::Result class that uses this. 

If we do want the DB to store 0/1, we could alter the `get_freezer` sub in `DBIx::Class::InflateColumn::Serializer::Boolean::JSON` to do it here.

drgrice1:
For the boolean database fields (those with `data_type => 'boolean'` in the schema, what you have done here is good.  That is not what i am talking about.  I am talking about the fields that have `data_type => 'text'` and use the `serializer_class => 'JSON'`.  For those fields, if the data that is store goes in as JSON true/false, then it comes out as a `Mojo::JSON::true` or `Mojo::JSON::false (which is really just `JSON::PP::Boolean`).  So the real problem there is that the data is NOT going in as JSON true/false if it is not coming back out right.  That is where this pull request goes wrong.  This hacks in something to do what the DBIx::Class::InflateColumn::Serializer::JSON really already does, and does in a more proper way.

The point is that by making it so that you need to name the boolean values inside the JSON column, you are limiting the versatility and expandability of those columns.  That was the real point of having those columns in the first place.

pstaabp:
The idea is that we'd like the database or Mojolicious layer to pass back proper JSON boolean values (not 0/1), so we can expect that in api requests and not have to parse things into javascript/typescript booleans. 

With discussion with @drdrew42, we thought that the technique is this current version of this PR might be the best way to do this.  Alternatively, if you are suggesting we keep the `DBIx::Class::InflateColumn::Serializer::JSON` inflators and then booleans are 0/1 in the DB, then in each of the `Schema::ResultSet` classes or at the Mojolicious layer, we could convert to `JSON::PP::Boolean` for appropriate fields.  

It seems like this code belongs in the `Schema::Result::` class level and when using ` DBIx::Class::InflateColumn::Serializer::JSON` there was no opportunity to convert boolean to `JSON::PP::Boolean` types.

drgrice1:
You still don't seem to be getting it.  If the data values are properly passed in as boolean values (not as 0's or 1's), then they will also be properly passed out as boolean values (not 0's or 1's).  DBIx::Class::InflateColumn::Serializer::JSON does this correctly, contrary to what you are saying.

pstaabp:
I understand now.  I didn't realize that ` DBIx::Class::InflateColumn::Serializer::JSON` did the boolean conversion correctly.   I'll drop the version of this that I wrote and double check everything.  

I just noticed that the `build_db.pl` script should be filling the params boolean fields correctly and it is not--this is perhaps why I took the more convoluted approach.

pstaabp:
@drgrice1 This is ready to take a look at again. The version of the inflator that I wrote is removed, the `build_db.pl` code is updated to write booleans into the params fields in the db and had to update a couple of tests.

pstaabp:
Actually, hold off for a bit on this.  I'm adding some tests to ensure that booleans inside of params fields are passed in as booleans, that is satisfies `JSON::PP::is_bool`.

pstaabp:
This time, we're ready to be tested again. Added checks to ensure that booleans are JSON booleans.

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -142,7 +142,10 @@ sub addCourse ($self, %args) {
 	}
 	$params->{course_settings} = {};
 
-	# Check the parameters.
+	# Add the visible default if not there.  Note: although in the database this value is
+	# already set, it appears that in the create method below, the default value is
+	# not created.  Perhaps rethink this a bit.
+	$params->{visible} = 1 unless defined $params->{visible};
 	my $new_course = $self->create($params);
```
drgrice1, line 149: This definitely needs to be rethought out.  The default value for the column in the database is set even if the 'visible' key is not in the params hash.  You are only adding this here so that it will be in the returned data.  Not because it has any effect at all on the creation of the course in the database.

What is really your intent with taking the return value of the `create` method, and using it for the return value of the `addCourse` method?  The only thing in that return value that has any real meaning is the `course_id`.

lib/DB/Schema/Result/ProblemSet/ReviewSet.pm
```diff
@@ -43,7 +43,7 @@ subroutine that returns the hashref for the valid parameters:  (currently empty)
 =cut
 
 sub valid_params ($=) {
-	return { test_param => qr/[01]/ };
+	return { test_param => 'bool' };
```
drgrice1, line 46: What is a review set `test_param`?  Is this just something you added for testing purposes?  This and all of the other related code should be deleted.  Implement actual code that will be used for webwork 3.

drgrice1 (changes requested):
This is a start of a review.  The database code needs to be though out better.  There is a lot of junk going on that really doesn't make sense in the code.

lib/DB/Schema/ResultSet/Course.pm
drgrice1: I think that the right way to handle this is to add `retrieve_on_insert => 1` to the visible column definition in the schema result source.  Then you can delete the line 148 (and preceding comments), and you will get the `visible` value back even if it is not in the params.

lib/DB/TestUtils.pm
```diff
@@ -25,33 +27,52 @@ field starting with PARAM: and DATE: respectively.
...
+			if ($key =~ /DATES:/) {
+				# Determine if each field is a date, a datetime or other (currently a boolean).
+				if (defined($input->{$key}) && $input->{$key} =~ /^\d{4}-\d{2}-\d{2}$/) {
+					my $dt = $strp_date->parse_datetime($input->{$key});
+					$output->{$field}->{$subfield} = $dt->epoch;
+				} elsif (defined($input->{$key}) && $input->{$key} =~ /^\d{4}-\d{2}-\d{2}T\d\d:\d\d:\d\dZ$/) {
+					my $dt = $strp_datetime->parse_datetime($input->{$key});
+					$output->{$field}->{$subfield} = $dt->epoch;
+				} elsif (grep {/^$subfield$/} @{ $config->{param_boolean_fields} }) {
+					$output->{$field}->{$subfield} = int($input->{$key}) ? true : false if defined($input->{$key});
```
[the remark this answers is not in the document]
pstaabp: In #89, I move the enable_reduced_scoring from the params field to the dates field and I think I moved this method from that branch to this one.

lib/DB/TestUtils.pm
```diff
@@ -25,33 +27,52 @@ field starting with PARAM: and DATE: respectively.
...
+				$output->{$field}->{$subfield} = 0 + $input->{$key} if defined($input->{$key});
+			} else {
+				$output->{$field}->{$subfield} = $input->{$key} if defined($input->{$key});
+			}
+		} elsif (grep {/^$key$/} @{ $config->{boolean_fields} }) {
+			$output->{$key} = defined($input->{$key}) && int($input->{$key}) ? true : false;
+		} elsif (grep {/^$key$/} @{ $config->{non_neg_int_fields} }) {
+			$output->{$key} = int($input->{$key}) if defined($input->{$key});
+		} elsif (grep {/^$key$/} @{ $config->{non_neg_float_fields} }) {
+			$output->{$key} = 0 + $input->{$key} if defined($input->{$key});
```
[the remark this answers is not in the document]
pstaabp: done.

src/common/models/user_sets.ts
```diff
@@ -188,10 +188,10 @@ export class DBUserHomeworkSet extends DBUserSet {
  */
 
 export type ParseableDBUserQuiz  = ParseableDBUserSet &
-{
+	{
 	set_params?: ParseableQuizParams;
 	set_dates?: ParseableQuizDates;
```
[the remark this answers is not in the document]
pstaabp: This may be a glitch. In the code, there are two tabs on lines 192, 193 and one tab on 191, 194.

lib/DB/Schema/Result/Course.pm
```diff
@@ -67,6 +67,14 @@ __PACKAGE__->add_columns(
 	}
 );
 
+__PACKAGE__->inflate_column(
+	'visible',
+	{
+		inflate => sub { return shift ? Mojo::JSON->true : Mojo::JSON->false; },
```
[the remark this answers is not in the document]
pstaabp: I agree.  Not sure why I did that. Good catch.

lib/DB/Schema/ResultSet/ProblemSet.pm
```diff
@@ -339,6 +339,9 @@ sub addProblemSet {
 		delete $set_params->{$key} if defined $set_params->{$key};
 	}
 
+	# set_visible should default to  false
+	$set_params->{set_visible} = 0 unless defined($set_params->{set_visible});
+
```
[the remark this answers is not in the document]
pstaabp: done.

t/db/001_courses.t
```diff
@@ -90,7 +92,9 @@ my $new_course_params = {
 	course_dates => {}
 };
 
-my $new_course      = $course_rs->addCourse(params => $new_course_params);
+my $new_course = $course_rs->addCourse(params => $new_course_params);
+# set the default value of visible
+$new_course_params->{visible} = 1;
```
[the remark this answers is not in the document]
pstaabp: yep.

lib/DB/Schema/Result/ProblemSet/ReviewSet.pm
pstaabp: This was just a dummy param for testing.  I'll remove it.

lib/DB/Schema/ResultSet/ProblemSet.pm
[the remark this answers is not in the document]
pstaabp: Actually, if we have "{'set_visible' => 0}" as the default, if the hash being set doesn't have the `set_visible` field, then it won't have this value.  I think I still need line 343, although it should be a JSON::false instead.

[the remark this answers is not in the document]
pstaabp: added the `retrieve_on_insert => 1` to set_visible.

lib/DB/Schema/Result/ProblemSet/ReviewSet.pm
pstaabp: It was a pain to remove it, so I renamed it to `can_retake` thinking that we can allow users to retake a review set.

drgrice1: For now that is okay.  I don't think that is a valid option either though.  If can_retake is false, then it should be a quiz.  Not a review set.

pstaabp: We can talk about what defines a review set, but the idea was that it doesn't count for a grade and perhaps solutions are always available.

lib/DB/Schema/ResultSet/ProblemSet.pm
```diff
@@ -9,6 +9,7 @@ use base 'DBIx::Class::ResultSet';
 
 use Carp;
 use Clone qw/clone/;
+use Mojo::JSON qw/true false/;
```
drgrice1, line 12: I don't think that these imports are used anywhere in this file.  So this should be removed.

lib/DB/TestUtils.pm
```diff
@@ -25,33 +27,52 @@ field starting with PARAM: and DATE: respectively.
...
+				# Determine if each field is a date, a datetime or other (currently a boolean).
+				if (defined($input->{$key}) && $input->{$key} =~ /^\d{4}-\d{2}-\d{2}$/) {
+					my $dt = $strp_date->parse_datetime($input->{$key});
+					$output->{$field}->{$subfield} = $dt->epoch;
+				} elsif (defined($input->{$key}) && $input->{$key} =~ /^\d{4}-\d{2}-\d{2}T\d\d:\d\d:\d\dZ$/) {
+					my $dt = $strp_datetime->parse_datetime($input->{$key});
+					$output->{$field}->{$subfield} = $dt->epoch;
+				} elsif (grep { $_ eq $subfield } @{ $config->{param_boolean_fields} }) {
+					$output->{$field}->{$subfield} = int($input->{$key}) ? true : false if defined($input->{$key});
+				}
```
drgrice1, line 47: This case should be removed from here, as I pointed out before.

src/common/models/user_sets.ts
```diff
@@ -564,11 +563,11 @@ export function parseUserSet(user_set: ParseableUserSet) {
 }
 
 /**
- * Merge a ProblemSet, a DBUserSet and a CourseUser and return a UserSet. Note: if the
+ * Merge a ProblemSet, a DBUserSet and a User and return a UserSet. Note: if the
  * arguments are not related in the database (based on primary and foreign keys), a MergeError is
- * thrown.  in that the result is a MergedUserSet with overrides taken from the UserSet.
+ * thrown.  The result is a UserSet with overrides taken from the UserSet.
```
drgrice1, line 568: This comment is confusing.  The UserSet is overridden by the same UserSet?  Looking at the code it seems that the default values from the `set.set_params` are used and those are overriden by values from the `db_user_set`.  Although it is a bit odd to use the terminology of override in this situation.  It seems that the effect is that the users values are used and the missing values are filled from the defaults.  So I would say `The result is a UserSet created with values from the DBUserSet and missing values taken from the ProblemSet.`

t/db/001_courses.t
```diff
@@ -90,7 +92,8 @@ my $new_course_params = {
 	course_dates => {}
 };
 
-my $new_course      = $course_rs->addCourse(params => $new_course_params);
+my $new_course = $course_rs->addCourse(params => $new_course_params);
+# set the default value of visible
```
drgrice1, line 96: This comment is no longer pertinent.

t/db/001_courses.t
```diff
@@ -36,13 +36,15 @@ my $strp   = DateTime::Format::Strptime->new(pattern => '%F', on_error => 'croak
...
 
 # Get a list of courses from the CSV file.
-my @courses = loadCSV("$main::ww3_dir/t/db/sample_data/courses.csv");
+my @courses = loadCSV(
+	"$main::ww3_dir/t/db/sample_data/courses.csv",
+	{
+		boolean_fields => ['visible']
+	}
+);
+use Data::Dumper;
```
drgrice1, line 45: Forgotten Data::Dumper to be deleted!

t/db/build_db.pl
```diff
@@ -115,27 +124,38 @@ sub addUsers {
...
+			param_boolean_fields => [ 'enable_reduced_scoring', 'hide_hint' ]
+		}
+	);
 	for my $set (@hw_sets) {
 		my $course = $course_rs->find({ course_name => $set->{course_name} });
 		if (!defined($course)) {
 			croak 'The course ' . $set->{course_name} . ' does not exist';
 		}
-
+		$set->{set_params} = {} unless defined $set->{set_params};
```
drgrice1, line 143: There is no need to set this.  The default will be filled in by the database if it is not defined.

t/db/build_db.pl
```diff
@@ -144,18 +164,26 @@ sub addSets {
 
 		$quiz->{type} = 2;
 		delete $quiz->{course_name};
+		$quiz->{set_params} = {} unless defined $quiz->{set_params};
```
drgrice1, line 167: Same here as above.

t/db/build_db.pl
```diff
@@ -144,18 +164,26 @@ sub addSets {
...
+			param_boolean_fields => ['can_retake']
+		}
+	);
 	for my $set (@review_sets) {
 		my $course = $course_rs->find({ course_name => $set->{course_name} });
 		croak "The course |$set->{course_name}| does not exist" unless defined($course);
 
 		$set->{type} = 4;
 		delete $set->{course_name};
+		$set->{set_params} = {} unless defined $set->{set_params};
```
drgrice1, line 186: And again. Not needed.

t/db/build_db.pl
```diff
@@ -214,6 +256,7 @@ sub addUserSets {
 			}
 			$user_set->{course_user_id} = $course_user->course_user_id;
 			$user_set->{set_id}         = $problem_set->set_id;
+			$user_set->{set_params}     = {} unless defined $user_set->{set_params};
```
drgrice1, line 259: Still not needed.

drgrice1 (commented):
There are a few more things that should be cleaned up.  I think this covers everything.

t/db/build_db.pl
pstaabp: done.

t/db/build_db.pl
pstaabp: done.

t/db/001_courses.t
pstaabp: done.

t/db/001_courses.t
pstaabp: done.

lib/DB/Schema/Result/ProblemSet/ReviewSet.pm
pstaabp: Agreed and fixed.

t/db/001_courses.t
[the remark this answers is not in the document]
pstaabp: done.

lib/DB/TestUtils.pm
pstaabp: done.

src/components/instructor/ClasslistManagerComponents/AddUsersFromFile.vue
```diff
@@ -309,7 +309,7 @@ const addMergedUsers = async () => {
 		let global_user: User | undefined;
 		try {
 			// Skip if username is undefined?
-			global_user = await getUser(user.username ?? '') as User;
+			global_user = await getUser(user.username ?? '') ;
```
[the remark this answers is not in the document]
pstaabp: fixed.

pstaabp:
Made those other changes.

src/components/instructor/ClasslistManagerComponents/AddUsersFromFile.vue
```diff
@@ -333,7 +333,7 @@ const addMergedUsers = async () => {
 		}
 		try {
 			await users.addCourseUser(new CourseUser(user));
-			const full_name = `${user.first_name as string} ${user.last_name as string}`;
+			const full_name = `${user.first_name as string } ${user.last_name as string}`;
```
drgrice1, line 336: Now just remove the added space, and put this file back to the way it was in the `main` branch and this pull request is ready.

pstaabp: Did you find this by eye? `npm run lint` or `quasar dev` doesn't catch these.  Or this something that falls under prettier instead of eslint?

drgrice1:
I found it by "eye".  We need to add the eslint rule `'template-curly-spacing': ['error', 'never']`.  That would automatically fix this.  In fact that is the only instance in the current code that breaks that rule.

pstaabp:
Good for merge now?

drgrice1 (approved):
Yeah, its good now.  Merging.

PR #91, opened 2022-06-01
run test-stores now spawns  morbo. 

pstaabp:
This changes the way that `npm run test-stores` runs.  Instead of requiring morbo to be started, this will spawn it before running the jest script. 

Additionally, morbo is started in 'test' mode which loads `conf/ww3-dev.yaml` instead of `conf/webwork3.yaml`. 

Other thoughts about this.  I set the port this runs as 3333.  Perhaps have an environment variable to set this.

Also, I haven't put this in the GitHub action. It might run now that morbo is running.


drdrew42 (approved):
Tested -- works even when ww3 backend is already running. Also confirmed that settings in ww3-dev are used, so no more swapping `cookie_secure` back and forth for testing!

drdrew42:
Dunno why this went from 5 changed files to 86, but I presume it's the usual GitHub UI inconsistency.

pstaabp:
Doing a git diff (via gitkraken) between main and this branch shows 4 changed files.

drgrice1:
If you merge main into this branch it will fix the issue.  You shouldn't have merged your model-improvement branch into this one.

drgrice1 (approved):
This is good enough for now.  The start of the `startup` method in `lib/WeBWorK3.pm` could use some reorganization in the future.

PR #93, opened 2022-06-17
FIX: update all of the quotes to single quotes unless interpolated.

pstaabp:
This switches all non-interpolated strings to single quotes.

### DIFF
--- a/bin/dev_scripts/PODParser.pm
+++ b/bin/dev_scripts/PODParser.pm
@@ -48,11 +48,11 @@ sub resolve_pod_page_link {
 	if ($self->{pod_found}) {
 		my $pod_url = $self->{pod_found} =~ s/^$self->{source_root}/$self->{base_url}/r;
 		undef $self->{pod_found};
-		print "Resolved local pod link $to" . ($section ? "/$section" : "") . " to $pod_url\n" if $self->{verbose} > 2;
+		print "Resolved local pod link $to" . ($section ? "/$section" : '') . " to $pod_url\n" if $self->{verbose} > 2;
 		return $self->encode_entities($pod_url);
 	}
 
-	print "Using cpan pod link for $to" . ($section ? "/$section" : "") . "\n" if $self->{verbose} > 2;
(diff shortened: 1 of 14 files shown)


bin/dev_scripts/PODParser.pm
```diff
@@ -48,11 +48,11 @@ sub resolve_pod_page_link {
 	if ($self->{pod_found}) {
 		my $pod_url = $self->{pod_found} =~ s/^$self->{source_root}/$self->{base_url}/r;
 		undef $self->{pod_found};
-		print "Resolved local pod link $to" . ($section ? "/$section" : "") . " to $pod_url\n" if $self->{verbose} > 2;
+		print "Resolved local pod link $to" . ($section ? "/$section" : '') . " to $pod_url\n" if $self->{verbose} > 2;
 		return $self->encode_entities($pod_url);
 	}
 
-	print "Using cpan pod link for $to" . ($section ? "/$section" : "") . "\n" if $self->{verbose} > 2;
+	print "Using cpan pod link for $to" . ($section ? "/$section" : '') . '\n' if $self->{verbose} > 2;
```
drgrice1, line 55: `\n` needs to be in a double quoted string.  `'\n'` does not create a newline.

bin/import_ww2_db.pl
```diff
@@ -6,7 +6,7 @@
 # This program is free software; you can redistribute it and/or modify it under
 # the terms of either: (a) the GNU General Public License as published by the
 # Free Software Foundation; either version 2, or (at your option) any later
-# version, or (b) the "Artistic License" which comes with this package.
+# version, or (b) the 'Artistic License' which comes with this package.
```
drgrice1, line 9: This should still be a double quote.  It is not a perl string that will be interpolated, but is part of a comment.  The correct copyright text should have double quotes for this.

drgrice1 (changes requested):
There are couple of minor issues.  Fix those and we can merge this.

drgrice1 (approved):

PR #94, opened 2022-06-17
Add two new perltidy options.

drgrice1:
First add the extended continuation indentation option. I added this to
the webwork2 and pg .perltidyrc files, and it does a better job of
indenting for nested structures.  This rule responsible for the changes
to the other files in this pull request.

Second add the option to disable vertical alignment of qw quotes.  Since
this option is added, the workflow version of perltidy can now be
allowed to increase.  I left the lock though, and just moved it to the
newest version.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -81,7 +81,7 @@ jobs:
       - name: perl -V
         run: perl -V
       - name: Install dependencies
-        run: cpanm -n Perl::Tidy@20220217
+        run: cpanm -n Perl::Tidy@20220613
       - name: perltidy --version
         run: perltidy --version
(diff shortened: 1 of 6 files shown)


drgrice1:
Actually, you don't need to. With the qw option added, there is no difference in the result.

pstaabp:
And the only files affected by the `xci` option are those in the PR. Correct?

drgrice1:
Yep.

pstaabp (approved):
Looks good and I think the additional nesting is easier to read.

drgrice1:
And the change is an improvement in those cases in my opinion also. That is why I had added it to the WeBWorK and pg rules.

PR #95, opened 2022-06-21
Redo roles and permissions

pstaabp:
This branch does a number of things with roles and permissions: 
- all roles and permissions are defined in `permissions.dist.yml`
- admins can copy this file over and change permissions/add roles, etc. 
- there is a role called `course_admin`, which can have additional permissions that instructors don't have. 
- there are both server-side permissions and ui permissions.
- there is a script `bin/update_perms.pl` which loads both the roles and permissions in the `permissions.yml` file into the database.
- on the server side, the permissions are checked by querying the database. 
- on the UI side, after login, the permissions are loaded in the store to be able to be queried whenever. 
- every route is now checked that the user has permission to visit that route. 

Some things that still need to be addressed:
- I branched this off main and there are some more tests in `t/mojolicious` that are in #88 and #89 that will need to be checked. 
- Wondering if this is sufficient or do we need another level of permissions similar to ww2 that is more fine grained?  If so, I think this can be added in a similar manner. 
- There's still some issues during testing, so if a user doesn't have permission, an error is logged, the user is not prevented from going to that page.  Clearly, this needs to be updated before being merged. 
- The routes on the ui-side are crude regexps and we may want to put the regexp in the permissions file.


drdrew42:
On first glance, I'm not a fan of swapping out the UserRole.unknown for empty string. If we're ENUM-ing the roles as we are, there needs to be a role for those who aren't logged in. Maybe change 'unknown' to 'notloggedin' or something indicative of the situation it is meant to represent. This serves to make the code more readable.

I'm also not clear on the new role: 'course_admin' ... what is this for? What will this provide for that the admin flag on the User doesn't already provide?

Also, it looks like you're building a many-to-many relationship (roles <-> actions) through a JSON field that holds an array. This should be a mapping table instead.

pstaabp:
This is ready to be tested now.  
* rebuild database (important as the roles/permissions are now stored in the database)
* run perl tests
* run typescript tests
* test UI (no new functionality, but some needed changes for permissions)

drdrew42:
Visit `/webwork3` without a cookie, view console:
* `[routing] The user does not have permission to visit '/login'`
* `[Vue Router warn]: The "next" callback was called more than once in one navigation guard when going from "/" to "/login". It should be called exactly one time in each navigation guard. This will fail in production.`

Log in as lisa, then logout, view console:
* `[Vue Router warn]: The "next" callback was called more than once in one navigation guard when going from "/users/3/courses" to "/login". It should be called exactly one time in each navigation guard. This will fail in production.`

Something is not being handled correctly in `index.ts` where the permission store is being put to use... it's working in dev mode, but these warnings need to be heeded -- this won't work in production.

pstaabp:
I merged this in.  I still get errors if not authenticated, but not the call to next() twice that you saw.  I think we discussed this earlier about more elegantly handling session expiring.  Was this supposed to handle going to /login?

drdrew42:
It certainly handles redirect to /login when logging out (which wasn't happening before -- we'd get the url for /login, but the page wasn't loading the login vue). It also handles redirection to the user_courses page when attempting to access any url for which the user does not have permission. I think we can extend this change to cover cookie timeout as well...

This does *not* address attempting to visit a url, being redirected to login, and afterwards maintaining the initial target url.

drgrice1:
@pstaabp: Sorry I didn't get to this today.  I should have time tomorrow though.

lib/DB/Schema/Result/DBPermission.pm
```diff
@@ -0,0 +1,96 @@
...
+	}
+);
+
+__PACKAGE__->inflate_column(
+	'admin_required',
+	{
+		inflate => sub { return shift ? Mojo::JSON->true : Mojo::JSON->false; },
+		deflate => sub { return shift; }
+	}
+);
```
drgrice1, line 74: It would be good to create an InflateColumn::Serializer class for these boolean columns instead of repeating this for every boolean column.  This is where the technique you used for the JSON columns would come in handy.

lib/DB/Schema/Result/User.pm
```diff
@@ -124,7 +124,7 @@ __PACKAGE__->inflate_column(
 __PACKAGE__->set_primary_key('user_id');
 __PACKAGE__->add_unique_constraint([qw/username/]);
 
-__PACKAGE__->has_many(course_users => 'DB::Schema::Result::CourseUser', { 'foreign.user_id' => 'self.user_id' });
+__PACKAGE__->has_many(course_users => 'DB::Schema::Result::CourseUser', 'user_id');
```
drgrice1, line 127: Why the change here?  These are equivalent, and what was wrong with the shorthand?

lib/DB/Schema/ResultSet/User.pm
```diff
@@ -272,9 +271,14 @@ or an arrayref of C<DBIx::Class::ResultSet::User>
...
-		'course_id' => $course->course_id
-	});
+	my @course_users = $self->rs('CourseUser')->search(
+		{
+			'course_id' => $course->course_id
+		},
+		{
+			prefetch => [qw/role/]
+		}
+	);
```
drgrice1, line 281: Why is this so spread out?  It can be done in two lines if you change it to
```suggestion
	my @course_users =
		$self->rs('CourseUser')->search({ 'course_id' => $course->course_id }, { prefetch => [qw/role/] });

```
pertidy accepts that.  The same can be done on lines 337 - 344 below.

lib/WeBWorK3/Controller/Permission.pm
```diff
@@ -0,0 +1,99 @@
...
+
+sub checkPermission ($c) {
+	# seems like $c->current_user doesn't have access to the course_id, so we never
+	# get a CourseUser.
+	my $user = $c->current_user;
+
+	# The controller/action pair is stored in the $c->match->stack array at the end.
+	# See https://docs.mojolicious.org/Mojolicious/Guides/Routing#Under
+	my $controller = $c->match->stack->[1]->{controller};
+	my $action     = $c->match->stack->[1]->{action};
```
drgrice1, line 44: There is unnecessary de-referencing here.  This should be
```suggestion
	my $controller = $c->match->stack->[1]{controller};
	my $action     = $c->match->stack->[1]{action};
```
Even the Mojolicious documentation you referenced does it this way.

There is more of this below in this file (and probably elsewhere).

@dpvc pointed this sort of thing out to me a few years ago.  webwork2 has massive amounts of this sort of unnecessary de-referencing.

lib/WeBWorK3/Controller/Problem.pm
```diff
@@ -4,87 +4,110 @@ use strict;
 
 use Mojo::Base 'Mojolicious::Controller', -signatures;
 
-sub addProblem ($self) {
-	my $problem = $self->schema->resultset('SetProblem')->addSetProblem(
-		params => {
-			course_id => int($self->param('course_id')),
-			set_id    => int($self->param('set_id')),
-			%{ $self->req->json }
+sub getAllProblems ($c) {
```
drgrice1, line 7: This conversion of `$self` to `$c` and `$self` to `$app` is exactly the sort of thing that you should do in a dedicated pull request.  If that is all that is in the pull request, it is easy to review and would be merged quickly.  By doing this in this pull request you make it so much harder to review, and slow down the merge of this pull request.

drgrice1 (commented):
This is just a start of my review.  I am out of time for now, and will continue it tomorrow.

lib/DB/Schema/Result/User.pm
drdrew42: isn't the changed version technically the 'shorthand'? When we have the same id on two tables, this is more concise?

drgrice1: Hah.  It seems that I read the diff in reverse.  The changed version is the shorthand.  Never mind then.

lib/WeBWorK3/Controller/Problem.pm
drdrew42: Yes, absolutely. But, to be clear, this is not a request to revert -- just a note for the future? These changes should be one cleanup PR, separate from a feature PR, yeah?

lib/WeBWorK3/Controller/Permission.pm
drdrew42: @pstaabp and I talked about this -- yes, we have absolutely picked up some bad habits from the ww2 codebase.

t/mojolicious/003_users.t
```diff
@@ -72,14 +75,37 @@ my $added_user_to_course = {
...
+# Check if a non-existent user is a global user
+$t->get_ok('/webwork3/api/courses/1/users/non_existent_user/exists')->status_is(200)
+	->content_type_is('application/json;charset=UTF-8');
+
+is_deeply($t->tx->res->json, {}, 'checkUserExists: check that a non-existent user returns {}');
+
+# Testing that booleans returned from the server are JSON booleans.
+# the first user is the admin
+
+my $admin_user = firstval { $_->{username} eq 'admin' } @all_users;
```
drgrice1, line 95: Something for the future:
I don't really like the usage of the `List::MoreUtils` module for only the `firstval` method as you have done all over in the tests and this is the only way this module is ever used.  That is something that can be done quite easily with native perl.  Just change this to
```suggestion
my $admin_user = ({ $_->{username} eq 'admin' } @all_users)[0];
```
That code is equivalent.  It returns the first element of the array, and is undefined if the array is empty.

If this is changed, that is one less module that we need as a dependency.

tests/unit-tests/utils.spec.ts
```diff
@@ -2,82 +2,84 @@
 
 import { pick, mapValues, invert, pickBy, random } from 'src/common/utils';
 
-test('testing pick', () => {
-	const obj = { a: 1, b: 2, c: 3, d: 4 };
+describe('Testing all utility functions', () => {
```
drgrice1, line 5: For the future again:
Surely you see that this change of adding descriptions is exactly the sort of completely unrelated change that should be done in another pull request.  These changes would even have almost no conflict with the other changes in this pull request, and if there are any minor conflicts they could easily be resolved.  This is what branches are for in git.

drgrice1 (changes requested):
This finishes my review of this pull request.  See my comments in the code for some things that should be changed, and some other things to think about for the future.

And the big show stopper ... running the quasar dev server and morbo, and opening localhost:8080 in the browser results in a blank screen, and webwork3 never loads.  There is the error "[error] [routing] User #0 does not have the permission to visit /login" in the console.  The user number can be different.  When testing in Firefox I got #0, and in Chrome I got #3.  However in both browsers the UI does not load.  In Firefox the browser hangs even, and it takes a while for it to finish loading.  This may be because I have the Vue dev tools installed there, but not in Chrome.

src/router/index.ts
```diff
@@ -28,5 +31,20 @@ export default route(() => {
 		),
 	});
 
-	return Router;
+	router.beforeResolve((to) => {
+		// must be logged in, or else headed to the login page
+		const session_store = useSessionStore();
+		if (!session_store.logged_in && to.name !== 'login') return { name: 'login' };
+
+		// otherwise, we're logged in, so check permissions on the route
```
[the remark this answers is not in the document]
drdrew42: This one is on me, I think. This all gets ripped out in an upcoming PR to handle session/cookie timeout. I'll work on eliminating the royal 'we' 👑  in the future.

src/stores/users.ts
```diff
@@ -135,7 +135,7 @@ export const useUserStore = defineStore('user', {
 		 * Updates the given user in the database and in the store.
 		 */
 		async updateUser(user: User): Promise<User | undefined> {
-			if (!user.isValid()) return invalidError(user, 'The updated user is invalid');
+			if (!user.isValid()) await invalidError(user, 'The updated user is invalid');
```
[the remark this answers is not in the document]
drdrew42: Yes. I'm guessing this is stemming from confusion over promises? `updateUser` intends to return a promise. On an error, however, it seems that `invalidError` is here to put a promise wrapper around the Error. Then, the `updateUser` caller expects a promise and should already be `await`ing (so it's incorrect to await here...)

src/layouts/MainLayout.vue
```diff
@@ -35,10 +35,11 @@ import MenuBar from './MenuBar.vue';
 
 import { instructor_views, admin_views, ViewInfo, student_views } from 'src/common/views';
 import { useRoute } from 'vue-router';
+import { logger } from 'boot/logger';
+
```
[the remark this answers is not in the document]
pstaabp: I'll be more careful with this. This looks like I just did some reorganization, but clearly, it doesn't belong here.

t/db/build_db.pl
```diff
@@ -265,7 +280,12 @@ sub addUserSets {
 
 sub addProblemPools {
 	say 'adding problem pools' if $verbose;
-	my @problem_pools = loadCSV("$main::ww3_dir/t/db/sample_data/pool_problems.csv");
+	my @problem_pools = my @problem_pools_from_file = loadCSV(
+		"$main::ww3_dir/t/db/sample_data/pool_problems.csv",
+		{
+			non_neg_int_fields => ['library_id']
+		}
+	);
```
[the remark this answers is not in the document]
pstaabp: I'll move over to this style.

t/mojolicious/003_users.t
pstaabp: Thanks for this.  I didn't know about this.

tests/unit-tests/utils.spec.ts
pstaabp: Got it.

t/mojolicious/003_users.t
pstaabp: Actually, don't we need a grep here, that is:
```
my $admin_user = (grep { $_->{username} eq 'admin' } @all_users)[0];
```

pstaabp:
I just did the recommended fixes.  A couple of notes:
- I created an inflator class to handle to the boolean--great idea--now you can just have a column type called 'json_bool' that just handles it. 
- changed all roles to lowercase.
- suggested code cleanup including removing `List::MoreUtils` and `firstval`

I didn't look into the UI.  I think that @drdrew42 session handling PR should fix this if we're okay with skipping fixing this.

drdrew42:
Once this is approved + merged, I have a fix for maintaining session duration in state, as well as route guarding and preserving the target URL until after login is completed.

drdrew42:
there is now a preview of my changes at pstaabp/webwork3#29

drdrew42:
I think changing capitalization of roles has more instances that need to be dealt with... e.g. logging in with lisa displays no courses -- they're in the store, but have roles in all caps still.

drgrice1:
> I just did the recommended fixes. A couple of notes:
> 
>     * I created an inflator class to handle to the boolean--great idea--now you can just have a column type called 'json_bool' that just handles it.
This isn't working at all.  I am a little surprised that sqlite works with this, but this completely fails with mysql and postgres.  I suggest that in the future you test all database changes with both.  Just run `./bin/setup_db.pl` with `webwork3.yml` set for either mysql or postgres to see what I mean.  The problem is that `json_bool` is not a valid database type.  You note in the `lib/DBIx/Class/InflateColumn/JSONBool.pm` file that it is a simplified version of `DBIx::Class::InflateColumn::DateTime`.  Note that the reason that package works is because `date` and `datetime` are valid database types.

In testing this I found another issue.  Data type `text` can not be used in a unique constraint.  You did this in `lib/DB/Schema/Result/Role.pm`.  The data type of the `role_name` column will need to be switched to `varchar`.

I put a pull request into your branch for this pull request that fixes these database issues.
> 
>     * changed all roles to lowercase.
> 
>     * suggested code cleanup including removing `List::MoreUtils` and `firstval`
This is certainly not finished.  You changed one instance (the one I specifically commented on).  The point is to entirely remove `firstval` from the code and thus the dependence on `List::MoreUtils`.  Also, my comment was a note for the future.  So don't do this in this pull request.  That should be done in a future pull request that is dedicated to doing only that.
> 
> 
> I didn't look into the UI. I think that @drdrew42 session handling PR should fix this if we're okay with skipping fixing this.
I am certainly not okay with merging a pull request that makes the UI completely dis-functional.

drdrew42:
So should we keep pstaabp/webwork3#29 open for review before we merge it into this branch?

drdrew42:
I will also switch over to postgres so that I'm testing with a more production-like environment.

drgrice1:
> So should we keep [pstaabp#29](https://github.com/pstaabp/webwork3/pull/29) open for review before we merge it into this branch?

Is that the "session handling PR" that @pstaabp referred to?  I didn't realize that he was referring to a pull request that adds to this one.  If that fixes the UI issue, then once that is merged we can merge this one.  Yeah, let me look at that pull request before it is merged into this one.

drdrew42:
Yes, I rebased my 'session handling' changes off of the branch for this PR. (And I opened the PR to @pstaabp's fork so that it could be reviewed without all the other changes already seen here.)

drdrew42:
I also kept my changes separate in case we decided to squash the commits here.

drdrew42:
`build_db.pl` needs changes to run, in addition to pstaabp/webwork3#30:
- insert new line at L110 `my $student_role = $role_rs->find({ role_name => 'student' });`
- edit L117 (now L118) `$course->add_to_users($stud_info, { role_id => $student_role->role_id });`

drgrice1:
@drdrew42: pstaabp#29 looks good for now and fixes the UI issue.

I added the lines you suggested to my pull request to this branch.  Those are certainly needed to use the build_db.pl script with mysql or postgres, otherwise there is a failed foreign key constraint issue.  I hadn't tested using the build_db.pl script with either of those databases.  I just had run setup_db.pl and saw the other issues.  Thanks for pointing that out.

pstaabp:
I'll be honest, I've only been running test on sqlite.  I'll switch to others before committing code.  Switching over to mariadb/mysql I'm getting test failing.  At least some are on arrays in which the order is flipped. @drgrice1 are all the tests running for you on mysql and Postgres?

drgrice1:
@pstaabp:  First, I should point out that when I say mysql I really mean mariadb.  I forget that I am actually running mariadb, largely because other than the ubuntu package name mysql and mariadb are the same.  Both use the same command line command name and same service name.

I am seeing that many of the unit tests are failing with both mariadb and postgres.  I am starting to think that using sqlite is a bad idea.  It is to lenient and allows with many things that more proper databases do not.  I think that we are going to need to rework the unit tests to test with both mariadb and postgres.

pstaabp:
I'm running mariadb right now and all but one of the failing tests is on an array and sorting them before testing fixes them.  I am having an issue with course_user and getting a fail on a addCourseUser.  It seems to be an issue with the many-to-many bridge table in mariadb. 

If I can get these fixed, I'll make sure I test in Postgres too.

drdrew42:
### Test Revisions Needed
- [ ] add quote_names => to schema connect
- [ ] handle sorting in tests (TBD: extend sort support to API queries)
- [ ] when using add_to_$rel on a many-to-many relation, use second argument to add params to merge table row

pstaabp:
Made a number of changes to get the test to run on mysql.  @drdrew42 said he will try under Postgres (I'll try later as well). 

Main fixes are:
- sorting arrays that come out of the database, so comparisons are made on sorted arrays.
- for UserProblem on the status field, the data_type was set right.  I have a numeric of size 10,5 (5 decimal places). Tests need to compare to numbers as strings to 5 decimal places.

pstaabp:
The tests are failing under sqlite now.  Will find a better solution to these.

drdrew42:
tests 007, 010, and 012 in t/db/ still need `quote_names => 1` (do not trust search-and-replace without checking that every expected file has been touched?)

drgrice1:
At this point the tests are failing for me with all databases.  I tested with mariadb, postgres, and sqlite.

pstaabp:
> tests 007, 010, and 012 in t/db/ still need `quote_names => 1` (do not trust search-and-replace without checking that every expected file has been touched?)

Will update 007, 010 and 012.  

> At this point the tests are failing for me with all databases. I tested with mariadb, postgres, and sqlite.

I may need to rethink the tests (especially those involving comparing two decimal numbers).

pstaabp:
I'm seeing all tests pass on sqlite, mysql/mariadb and Postgres.

src/common/models/users.ts
```diff
@@ -187,7 +183,7 @@ export class CourseUser extends Model {
 	private _first_name = '';
 	private _last_name = '';
 	private _student_id = '';
-	private _role = UserRole.unknown;
+	private _role = 'UNKNOWN';
```
drgrice1, line 186: UNKNOWN should be lower case here.

src/common/models/users.ts
```diff
@@ -266,7 +258,7 @@ export class CourseUser extends Model {
 	isValid(): boolean {
 		return isNonNegInt(this.user_id) && isNonNegInt(this.course_user_id) &&
 			isNonNegInt(this.course_id) && isValidUsername(this.username) &&
-			this.role !== UserRole.unknown && (this.email === '' || isValidEmail(this.email));
+			this.role !== 'UNKNOWN' && (this.email === '' || isValidEmail(this.email));
```
drgrice1, line 261: UNKNOWN should be lower case here.

src/common/models/users.ts
```diff
@@ -276,7 +268,7 @@ export class CourseUser extends Model {
 			user_id: isNonNegInt(this.user_id) || 'The user_id must be a non negative integer.',
 			username: isValidUsername(this.username) || 'The username must be valid.',
 			email: (this.email === '' || isValidEmail(this.email)) || 'The email must be valid',
-			role: this.role !== UserRole.unknown || 'The role is not valid.',
+			role: this.role !== 'UNKNOWN' || 'The role is not valid.',
```
drgrice1, line 271: UNKNOWN should be lower case here.

drgrice1 (commented):
There are a few minor changes to the role casing that are still needed.

Line 198 of src/components/instructor/ClassListManagerComponents/AddUsersFromFile.vue is another instance of where UNKNOWN should be lower case.  Furthermore there is a mispelled UNKOWN on that line.

Line 143 of tests/unit-tests/course_users.spec.ts has another instance where UNKNOWN should be lower case.

I am not sure that I agree that there should be a `course_admin` role.  We should talk about this.

drgrice1:
Once the casing instances are fixed we can merge this.

pstaabp:
Just fixed these.  Noticed that the problem set types are also upper case.  I will change those in a separate PR. 

About the 'course_admin' role, I'm not wedded to it, but just building the infrastructure and if we have permissions set up right we should be able to change roles and what permissions they have in the conf file.

drgrice1:
Once the tests are complete I will merge this.

@drdrew42:  Are you okay with a squash merge?  Doing so will remove our commits, however the commit message in the new squashed commit will list us as co-authors.  Github automatically adds
```
Co-authored-by: K. Andrew Parker <kparker@citytech.cuny.edu>
Co-authored-by: Glenn Rice <grice1@missouriwestern.edu>
```
to the end.

@pstaabp:  I did notice the uppercase problem set types.

drgrice1:
> Line 143 of tests/unit-tests/course_users.spec.ts has another instance where UNKNOWN should be lower case.

This line didn't get changed.  That is why the tests failed.

pstaabp:
I'm fine with a squash merge.  Uggh.  Another fail!!

pstaabp:
And trying to be more focused with PRs and  I have figured out the rebasing--the latest two PRs I have have only the commits on those branches, so moving ahead, these should be cleaner.

drdrew42 (approved):
Let's squash

drgrice1:
Okay.  It is in.

PR #99, opened 2022-06-30
permissions problems related to UI

pstaabp:
There were a number of problems on the UI related to permissions problems.  

This should fix the problem with adding users either manually or via a file.


pstaabp:
I didn't update this after merging #88.  I think #89 should take care of these issues as well.

pstaabp:
The fixes in here are in #89.  Closing for now.

PR #100, opened 2022-07-08
Add the eslint rule `'template-curly-spacing': ['error', 'never']`.

drgrice1:
This enforces that there are no spaces inside curly brackes in a template literal, and these are automatically fixable when eslint is run!


pstaabp:
I'll merge this.

PR #102, opened 2022-07-20
Create an Instructor Dashboard

pstaabp:
This creates a very basic instructor dashboard (mainly with instructions to select a different tool) as well as fixes some errors when switching courses as a user.

This contains #89, so that should be merged before this.


pstaabp:
Closing this in favor of #109 in which main was rebased onto.

PR #105, opened 2022-07-22
Clean up the conf file handling.

drgrice1:
Rename ww3-dev.dist.yml to webwork3-test.dist.yml.  The ww3 abreviation is not needed and is inconsistent in naming with the other conf files.

You can now create a separate webwork3-dev.yml file for use in development.  There is no associated dist file for this.  Just make
whatever changes you want to either of the webwork3*.dist.yml files.

Delete the we_b_wor_k3.yml link.  Just load the appropriate conf file in conf directly.

Completely eliminate the WW3_ROOT environment variable.  There is no need for that.  Instead make use of Mojo::Home.  It is made for this. If you are in an area of the code that has access to the app, then just use the home helper (in a controller this can be accessed via $c->app->home).  Elsewhere in the code use Mojo::Home as is done in lib/WeBWorK3/Utils/Settings.pm.

Also fix the usage of the exception classes in lib/WeBWorK3/Utils/Settings.pm.  My editor complains about the way those
were imported there.  Furthermore it was entirely inconsistent with the way the exception classes are imported elsewhere.  There is a mess of ways that imports are done in the perl files.  This needs further clean up.  In some places qw// is used, in others qw() is used (both of those occur in lib/WeBWorK3/Utils/Settings.pm), in yet others a list of individually quoted imports is used (see lib/DB/WithParams.pm).

@pstaapb: Please find some consistency in your coding style.


bin/import_ww2_db.pl
```diff
@@ -81,7 +81,7 @@ BEGIN
 # Load the webwork3 configuration file.
 
 my $config_file = "$main::ww3_dir/conf/webwork3.yml";
-die "The file $config_file does not exist.  Did you make a copy of it from ww3-dev.dist.yml ?"
```
pstaabp, line 84: Should we use the same strategy here as in other files, where if webwork3.yml doesn't exist, we use webwork3.dist.yml?

drgrice1: Yeah, that sounds good.  I changed that now in this pull request.

pstaabp (approved):

PR #106, opened 2022-08-02
Rework the Course Settings

pstaabp:
This reworks the way course settings are handled on the back end.  Specifically
- store the global/default course settings in the database
- store the override course settings as a single row in a table (much simpler than the old method)

There are new routes for fetching/updating/deleting global and course settings.

Because of this, the settings modules on the front end side changes to reflect the back end changes.  Including
- inheriting from the Model class for both global settings and course settings (overrides)
- in the settings store, the global_settings and `db_course_settings` are stored separately (similar to other data in the stores) and then merged. 

There are a lot more tests now including
- `t/db/002_course_settings.t` is rewritten to reflect the new course settings back end database model.
- `t/mojolicious/015_course_settings.t` is a new to test all of the routes
- `tests/unit-tests/settings.ts` tests the new model and handling of exceptions.
- `tests/stores/settings.ts` tests the settings store.

The UI is mostly unchanged except for using some new store getters and now has the ability to save course settings. 

To test
- run `t/db/build_db.pl`
- `prove -r t`
- `npm run test`
- `npm run test-store`
- The UI--mainly the Settings view as an instructor.  Changes are sent immediately to the database and you should see a confirmation notification. 

Note: I think GitHub is not accurately managing the changes.  A git diff between this and main is 20 files, mostly related to settings.


pstaabp:
Closing this since #108 is identical and has been rebased onto main.

PR #108, opened 2022-08-06
Course settings refactoring

pstaabp:
This replaces #106 with better rebasing against main.

This reworks the way course settings are handled on the back end.  Specifically
- store the global/default course settings in the database
- store the override course settings as a single row in a table (much simpler than the old method)

There are new routes for fetching/updating/deleting global and course settings.

Because of this, the settings modules on the front end side changes to reflect the back end changes.  Including
- inheriting from the Model class for both global settings and course settings (overrides)
- in the settings store, the global_settings and `db_course_settings` are stored separately (similar to other data in the stores) and then merged. 

There are a lot more tests now including
- `t/db/002_course_settings.t` is rewritten to reflect the new course settings back end database model.
- `t/mojolicious/015_course_settings.t` is a new to test all of the routes
- `tests/unit-tests/settings.ts` tests the new model and handling of exceptions.
- `tests/stores/settings.ts` tests the settings store.

The UI is mostly unchanged except for using some new store getters and now has the ability to save course settings. 

To test
- run `t/db/build_db.pl`
- `prove -r t`
- `npm run test`
- `npm run test-store`
- The UI--mainly the Settings view as an instructor.  Changes are sent immediately to the database and you should see a confirmation notification. 

Note: I think GitHub is not accurately managing the changes.  A git diff between this and main is 20 files, mostly related to settings.


pstaabp:
I notice I should have posted as a draft.  The UI is mostly working now.  A couple things I'd like to do:
- [ ] switch any time_duration setting to a number of seconds in the database and let the UI handle the parsing of strings (like `15 mins` to 900) and vise versa.
- [ ] Make sure the permissions are working for fetch/put settings.

pstaabp:
This is ready to go now.  One question.  In the db for both global settings (defaults) and course settings (overrides), the value is stored as a JSON in the form `{value => ....}` because the value can be many different types.  to handle this, I manually inflate (insert and pull out of the hash)--it only happens in a few places, but should I do something more generally?   I haven't done that because it seems like it would be replacing the `JSON::Serializer` that is used now.  Thoughts?

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -254,6 +255,78 @@ sub getUserCourses ($self, %args) {
...
+	my @global_settings = $self->result_source->schema->resultset('GlobalSetting')->search({});
+
+	return \@global_settings if $args{as_result_set};
+	my @settings = map {
+		{ $_->get_inflated_columns };
+	} @global_settings;
+	for my $setting (@settings) {
+		# The default_value is stored as a JSON and needs to be parsed.
+		$setting->{default_value} = $setting->{default_value}->{value};
+	}
```
drdrew42: Instead of looping twice over the same list, why not combine the actions into a single loop?

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -262,36 +335,202 @@ This gets the Course Settings for a course
...
+		: map {
+			{ $_->get_inflated_columns };
+		} @settings_from_db;
+
+	for my $setting (@settings_to_return) {
+		# value and default_value are decoded from JSON as a hash.  Return to a value.
+		for my $key (qw/default_value value/) {
+			$setting->{$key} = $setting->{$key}->{value} if defined($setting->{$key});
+		}
+	}
```
drdrew42: Again, we're using two separate loops over the same data.

Also, it's unnecessary to use `->` more than once when traversing nested hashes.

Finally, what is going on in line 370? Replace a key-value if the key already has a defined value? Shouldn't this be `if defined($setting->{$key}{value})`?

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -262,36 +335,202 @@ This gets the Course Settings for a course
...
+		? { $setting->get_inflated_columns, $setting->global_setting->get_inflated_columns }
+		: { $setting->get_inflated_columns };
 
-	my $course_settings  = getDefaultCourseValues();
-	my $settings_from_db = { $course->course_settings->get_inflated_columns };
-	return mergeCourseSettings($course_settings, $settings_from_db);
+	# value and default_value are decoded from JSON as a hash.  Return to a value.
+	for my $key (qw/default_value value/) {
+		$setting_to_return->{$key} = $setting_to_return->{$key}->{value} if defined($setting_to_return->{$key});
+	}
```
drdrew42: It seems that we're doing this across multiple calls... why not create a util function that will do an in-place edit on a hash?
```perl
sub hash2value ($hash) {
    for (qw/default_value value/) {
        $hash->{$_} = $hash->{$_}{value} if defined($hash->{$_}{value});
    }
}
```

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -262,36 +335,202 @@ This gets the Course Settings for a course
...
+		value      => { value => $args{params}{value} }
+	};
+
+	# remove the following fields before checking for valid settings:
+	delete $global_setting->{$_} for (qw/setting_id course_id/);
+
+	isValidSetting($global_setting, $params->{value}{value});
+
+	# The course_id must be deleted to ensure it is written to the database correctly.
+	delete $params->{course_id} if defined($params->{course_id});
```
drdrew42: This is awfully confusing. Our setting validation subroutine requires some keys to be deleted, and others to be added, which we don't actually want to use? Maybe we should re-think how `isValidSetting` is implemented, because this looks bizarre (we're defining `$params`, apparently with key-values that we don't even want -- just so we can validate?)

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -262,36 +335,202 @@ This gets the Course Settings for a course
...
+	my $setting_to_return =
+		($args{merged})
+		? { $deleted_setting->get_inflated_columns, $deleted_setting->global_setting->get_inflated_columns }
+		: { $deleted_setting->get_inflated_columns };
+
+	# value and default_value are decoded from JSON as a hash.  Return to a value.
+	for my $key (qw/default_value value/) {
+		$setting_to_return->{$key} = $setting_to_return->{$key}->{value} if defined($setting_to_return->{$key});
+	}
+	return $setting_to_return;
```
drdrew42: I thought we discussed *not* returning objects from deletion calls...

drdrew42 (commented):
Initial read-through and feedback on the db side of this PR

lib/DB/Schema/ResultSet/Course.pm
pstaabp: This sounds good.  I also commented above on if this would be helpful to make an inflate/deflate function that does this.

lib/DB/Schema/ResultSet/Course.pm
pstaabp: I wrote this before the deleteXXX PR.  I can update.

pstaabp: done.

lib/DB/Schema/ResultSet/Course.pm
pstaabp: this is cleaned up.  I'm ignoring any invalid fields now (must like we are doing throughout the codebase).

lib/DB/Schema/ResultSet/Course.pm
pstaabp: The inflator cleans this up a lot.

lib/DB/Schema/ResultSet/Course.pm
pstaabp: Again, the inflator takes care of this.

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -400,25 +386,31 @@ A single course setting as either a hashref or a C<DBIx::Class::ResultSet::Cours
...
 		unless defined($global_setting);
 
 	my $course  = $self->getCourse(info => getCourseInfo($args{info}), as_result_set => 1);
 	my $setting = $course->course_settings->find({ setting_id => $global_setting->setting_id });
 
+	DB::Exception::SettingNotFound->throw(
+		message => 'The course setting with '
+			. (
+				$args{info}->{setting_name} ? " name: '$args{info}->{setting_name}'"
+				: "setting_id of $args{info}->{setting_id} is not a found in the course "
```
drdrew42: "is not ~~a~~ found"

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -400,25 +386,31 @@ A single course setting as either a hashref or a C<DBIx::Class::ResultSet::Cours
 =cut
 
 sub getCourseSetting ($self, %args) {
-
 	my $global_setting = $self->getGlobalSetting(info => $args{info}, as_result_set => 1);
 	DB::Exception::SettingNotFound->throw(
-		message => "The setting with name: '" . $args{info}->{setting_name} . "' is not a defined info.")
+		message => "The global setting with name: '" . $args{info}->{setting_name} . "' is not a defined info.")
```
drdrew42: What does this exception mean? _"is not a defined info"_

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -400,25 +386,31 @@ A single course setting as either a hashref or a C<DBIx::Class::ResultSet::Cours
...
 		unless defined($global_setting);
 
 	my $course  = $self->getCourse(info => getCourseInfo($args{info}), as_result_set => 1);
 	my $setting = $course->course_settings->find({ setting_id => $global_setting->setting_id });
 
+	DB::Exception::SettingNotFound->throw(
+		message => 'The course setting with '
+			. (
+				$args{info}->{setting_name} ? " name: '$args{info}->{setting_name}'"
```
drdrew42: `$args{info}{setting_name}` does not require any thin arrows. Thin arrows are only needed when de-referencing the initial HASH ref (or making method calls). This is a pattern I think we all picked up from working on ww2, it's not needed.

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -323,7 +319,6 @@ sub getGlobalSetting ($self, %args) {
 		unless $global_setting;
 	return $global_setting if $args{as_result_set};
 	my $setting_to_return = { $global_setting->get_inflated_columns };
-	$setting_to_return->{default_value} = $setting_to_return->{default_value}->{value};
```
drdrew42: we can ditch the line above this one as well and just `return { $global_setting->get_inflated_columns };`

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -451,44 +443,30 @@ A single course setting as either a hashref or a C<DBIx::Class::ResultSet::Cours
...
 
-	return $updated_course_setting if $args{as_result_set};
+	return $up_setting if $args{as_result_set};
 	my $setting_to_return =
 		($args{merged})
-		? { $updated_course_setting->get_inflated_columns,
-			$updated_course_setting->global_setting->get_inflated_columns }
-		: { $updated_course_setting->get_inflated_columns };
+		? { $up_setting->get_inflated_columns, $up_setting->global_setting->get_inflated_columns }
+		: { $up_setting->get_inflated_columns };
```
drdrew42, line 459: here we can also just return instead of creating `$setting to return`

drdrew42 (commented):
Just a few more cleanup things that arise as a result of simplifying with the new inflate method.
- places where we needed a variable, now we can just return instead
- avoid unnecessary `->` when traversing nested hashes

lib/DB/Schema/Result/CourseSetting.pm
```diff
@@ -0,0 +1,69 @@
...
+		is_auto_increment => 1,
+	},
+	course_id => {
+		data_type => 'integer',
+		size      => 16,
+	},
+	setting_id => {
+		data_type => 'integer',
+		size      => 16,
+	},
```
drgrice1: This is a bad name for this column, and the description above is equally poor.  At first glance it is unclear why `course_setting_id` and `setting_id` columns both exist in this table, and the description doesn't clear that up.  A better name would be `global_setting_id`, and a better description would be "database id of the global setting that the given setting is related to".

The primary key of the `global_setting table` should also be changed to be the `global_setting_id` as the current key is inconsistent with the primary key naming used in tables in general in what we have so far of the ww3 database.

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -262,36 +330,172 @@ This gets the Course Settings for a course
...
+	my $course           = $self->getCourse(info => $args{info}, as_result_set => 1);
+	my @settings_from_db = $course->course_settings;
+
+	return \@settings_from_db if $args{as_result_set};
+	my @settings_to_return = map {
+		$args{merged}
+			? { $_->get_inflated_columns, $_->global_setting->get_inflated_columns }
+			: { $_->get_inflated_columns };
+	} @settings_from_db;
+	return \@settings_to_return;
```
drgrice1: Just return the `map ...` statement.  There is no need for the `@settings_to_return` variable.  Change this to
```suggestion
	return [
		map {
			$args{merged}
				? { $_->get_inflated_columns, $_->global_setting->get_inflated_columns }
				: { $_->get_inflated_columns };
		} @settings_from_db
	];
```
It looks like there are many other places where this needs to be done also.  I commented on a few above, but there are several others in this file (and probably elsewhere in the code).  Don't create a `..._to_return` variable, and then immediately return it.  Just return the object that you assign to that variable.

This is something I told you about before, and applies to almost all languages.  Don't create variables that are used only once.  Just use whatever it is that you assign to that variable directly.

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -254,6 +255,73 @@ sub getUserCourses ($self, %args) {
...
+=cut
+
+sub getGlobalSettings ($self, %args) {
+	my @global_settings = $self->result_source->schema->resultset('GlobalSetting')->search({});
+
+	return \@global_settings if $args{as_result_set};
+	my @settings = map {
+		{ $_->get_inflated_columns };
+	} @global_settings;
+	return \@settings;
```
drgrice1, line 286: See my comment later in this file.  `@settings` is not needed.

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -254,6 +255,73 @@ sub getUserCourses ($self, %args) {
...
+	my $setting_info   = getSettingInfo($args{info});
+	my $global_setting = $self->result_source->schema->resultset('GlobalSetting')->find($setting_info);
+
+	DB::Exception::SettingNotFound->throw(message => $setting_info->{setting_name}
+		? "The setting with name $setting_info->{setting_name} is not found"
+		: "The setting with setting_id $setting_info->{setting_id} is not found")
+		unless $global_setting;
+	return $global_setting if $args{as_result_set};
+	my $setting_to_return = { $global_setting->get_inflated_columns };
+	return $setting_to_return;
```
drgrice1: Again, see later comment.

lib/DB/Schema/ResultSet/Course.pm
```diff
@@ -262,36 +330,172 @@ This gets the Course Settings for a course
...
+
+=cut
+
+sub updateCourseSetting ($self, %args) {
+	my $course         = $self->getCourse(info => getCourseInfo($args{info}), as_result_set => 1);
+	my $global_setting = $self->getGlobalSetting(info => getSettingInfo($args{info}));
+
+	my $course_setting = $course->course_settings->find({
+		setting_id => $global_setting->{setting_id}
+	});
```
drgrice1: This could be consolidated to one line.  It is just as readable, and it is always a good idea to make methods shorter by doing so.

lib/DB/Utils.pm
```diff
@@ -189,4 +193,55 @@ sub updatePermissions ($ww3_conf, $role_perm_file) {
...
+	my $time_duration = '';
+	# Order is important so the keys are defined.
+	for (qw/week day hour min sec/) {
+		my $val = $times->{$_};
+		$time_duration .= ($time_duration ne '' && $val ? ', ' : '')
+			# pluralize for more than 1.
+			. ($val > 0 ? "$val $_" . ($val == 1 ? '' : 's') : '');
+	}
+	return $time_duration;
+}
```
drgrice1, line 245: This should be deleted and dealt with client side.  The database (or the api in general) should not be dealing with anything related to "human readability".  The database should return a unix timestamp (or at least a ISO timestamp).  The client UI should handle converting that into a human readable thing.

lib/DBIx/Class/InflateColumn/JSONValue.pm
```diff
@@ -0,0 +1,45 @@
...
+					return $str =~ /^-?(0|([1-9][0-9]*))(\.[0-9]+)?([eE][-+]?[0-9]+)?$/ ? 0 + $str : $str;
+				};
+			},
+			deflate => sub { return encode_json({ value => shift }); }
+		}
+	);
+	return;
+}
+
+1;
```
drgrice1: NO.  Find a different way of accomplishing this.  This is conceptually bad, and is not a robust data storage design.  Absolutely not.  Ugh!  This poor design is part of what led to @drdrew42's comments on the `{default_value}{value}` in his review.

lib/WeBWorK3/Controller/Settings.pm
```diff
@@ -3,47 +3,89 @@ package WeBWorK3::Controller::Settings;
...
 =cut
 
 use Mojo::Base 'Mojolicious::Controller', -signatures;
-use Mojo::File qw/path/;
 
-use YAML::XS qw/LoadFile/;
+sub getGlobalSettings ($c) {
+	my $settings = $c->schema->resultset('Course')->getGlobalSettings();
+	$c->render(json => $settings);
+	return;
```
drgrice1: This file is filled with more examples of unnecessary variable declarations.  Just put `$settings` into the only place it is used.  This method can be made quite concise by changing it to
```suggestion
	$c->render(
		json => $c->schema->resultset('Course')->getCourseSettings(
			info   => { course_id => int($c->param('course_id')) },
			merged => 1
		)
	);
	return;
```
Repeat throughout this file.

drgrice1 (changes requested):
I have reviewed the API changes in this pull request.  At this point this pull request is no where close to being ready to go.  This needs to be completely redesigned.  The structure is convoluted and doesn't make any sense.

There is a huge problem with webwork3 in its entirety at this point.  Nothing is coming together into a coherent whole.  Every pull request makes things more structurally messy instead of unifying things.

The JSONValue.pm concept ... see my comments.  See @drdrew42's comments.  It is time for you to stop putting in pull requests so hastily.  Take some time and think things through.  Spend time testing them and developing better structure.

When the API portion of this is starting to look more reasonable, then I will look at the rest.

drgrice1:
I also am not convinced that the setup with both the permissions and course settings is a good idea.  Having a YAML file that has these settings that must then be loaded into the database will be a system administrative nightmare.  When you want to change a single setting for the site, then you have to modify the YAML file, then load it into the database.  If things go wrong with that, then the entire site is in trouble.  This will lead to an excessive need to constantly back up the database and deal with recovery in case of failure.

Instead it would make more sense for the YAML files to directly be the global defaults.  The database should then store overrides for a course.

lib/DBIx/Class/InflateColumn/JSONValue.pm
```diff
@@ -0,0 +1,45 @@
...
+
+	$self->next::method($column, $info, @rest);
+	return unless $info->{inflate_value};
+
+	$self->inflate_column(
+		$column => {
+			inflate => sub {
+				my $str = shift;
+				# This is a bit of a hack.  It appears that sometimes the deflate isn't called on the values
+				# of type string and number so they don't need to be decoded.
```
drgrice1: Did you think to try the `allow_nonref` option for the `JSON` serializer instead of this ugly hack?

lib/DB/Schema/ResultSet/Course.pm
pstaabp: I've been try to return arrayrefs from these functions and doesn't map just create an array?  This is why I did this.

lib/DB/Utils.pm
pstaabp: The reason I have these is to have human-readable in the configuration files and these subroutines are used to translate that.  Also, humanReadableTimeDuration is used in tests.  If we keep this, this should be moved to the t/lib directory.

lib/DBIx/Class/InflateColumn/JSONValue.pm
pstaabp: I had tried this.  It didn't seem to work, but right now, I don't recall why.  I will try using `allow_nonref` again.

pstaabp:
> Having a YAML file that has these settings that must then be loaded into the database will be a system administrative nightmare.

I was looking at this as similar to what we decided to do with the permissions, that is move them into the database.  We can talk about this on Wednesday.

lib/DB/Schema/ResultSet/Course.pm
drgrice1: That doesn't change anything with the point here.  No single use variables.  I fixed my suggestion so that an array ref is returned.

Edit: Now it returns an array ref (it was temporarily a hash ref with potentially an odd number of entries).

lib/DB/Schema/ResultSet/Course.pm
pstaabp: fixed.

lib/DB/Schema/ResultSet/Course.pm
pstaabp: fixed.

lib/DB/Schema/ResultSet/Course.pm
pstaabp: Fixed all of these.

lib/DB/Schema/ResultSet/Course.pm
pstaabp: fixed.

lib/DB/Schema/ResultSet/Course.pm
pstaabp: fixed.

lib/DB/Utils.pm
drgrice1: This is something that need to reconsidered even deeper.  What settings here are being saved in config files?  If a setting needs conversion to become readable then it probably is being done wrong.  Basically, it is probably incorrect to have a date at all in a config file.  Time durations can be purely in seconds (or minutes or whatever flat time unit) and still work perfectly well in a config file and be human readable.

drgrice1:
> > Having a YAML file that has these settings that must then be loaded into the database will be a system administrative nightmare.
> 
> I was looking at this as similar to what we decided to do with the permissions, that is move them into the database. We can talk about this on Wednesday.

Yeah, that is my whole point.  I listed permissions here, and I am saying that was a bad idea.  I was not involved in that decision, and if I was I would have told you to go a different direction.

lib/DB/Schema/ResultSet/Course.pm
drgrice1: Even better is to not check the `$args{merged}` value for each entry of the `@settings_from_db` array.  That is constant for the iteration, and so this should be
```perl
	return $args{merged}
		? [ map { { $_->get_inflated_columns, $_->global_setting->get_inflated_columns } } @settings_from_db ]
		: [ map { { $_->get_inflated_columns } } @settings_from_db ];
```

lib/WeBWorK3/Controller/Settings.pm
pstaabp: This looks good.  In a future PR I can change the other Controller modules to do this as well.

drgrice1: Note this comment was meant to be for lines 35-42 in this file.  Not lines 19-21.

drgrice1: Although the concept applies here too.

lib/DB/Utils.pm
pstaabp: My thought with this is that time durations, it's easier to think in terms of minutes or days often instead of numbers of seconds (although clearly easy to calculate), so providing this functionality is relatively simple.  Again, we can chat more about this.

pstaabp:
I don't think this is too late to switch without changing the API.

lib/DBIx/Class/InflateColumn/JSONValue.pm
pstaabp: I realized the trouble with using JSON and allow_nonref was that of parsing empty strings.  I've switched this back to using these two and working on handling empty strings.  This is definitely the better way to go.

drgrice1: What is the issue with parsing empty strings?  Also, what do you mean you switched bak to using "these two"?  What two?  And, what are you saying is definitely the better way to go?

pstaabp: Sorry for the confusion. I'll try to be more clear.  I think this was all an issue with the InflateColumn package. I have switched over to using FilterColumn, which looks like the way to go.

lib/DB/Schema/ResultSet/Course.pm
pstaabp: This is what I needed. Switched to this.

pstaabp:
Made all of the requested changes.  The inflator for settings (either default_value or value) is now using FilterColumn, which has more flexibility over InflateColumn. 

One argument for keeping the settings (and permissions) in the database is that they are checked before adding to the database, ensuring the types are correct.  If they in a file, it would be simply parsed or checked on each file read.

drgrice1:
The problem is that you are thinking we need yml files at all for this.  Better would be a administrative user interface for editing roles, and no yml files at all.

drgrice1:
For course settings there are already big problems with the yml file approach.  The `description` and `doc` should not be in the yml file or database at all.  Those should be implemented in the client ui.  Part of the problem is that if those are API side, they will not work with translations.  Having these in the config files was a huge mistake made for webwork2 to begin with not only for translations.  Those are things that really should never be configurable.  If they are in defaults.config, then a config file loaded later can override them, and that shouldn't be possible.  For webwork2 they should have been in the code and added to the hash objects later.  Hard coded and not user or even sys admin modifiable.

pstaabp:
>  The `description` and `doc` should not be in the yml file or database at all. Those should be implemented in the client ui. Part of the problem is that if those are API side, they will not work with translations. 

Great point and should either include a translation key or perhaps create a key from the setting_name. 

Your other points are also good to think about moving forward.  It is convenient at this point of having a single file with all of the settings.  I think the settings will adapt quite a bit as we develop ww3.  My hope was to have a flexible system to handle this, but I understand after we have nailed this down, allowing overrides for global settings provides difficulties.

PR #109, opened 2022-08-06
Create a basic instructor dashboard

pstaabp:
This creates a very basic instructor dashboard (mainly with instructions to select a different tool) as well as fixes some errors when switching courses as a user.

This replaces #102.  Main was rebased onto this branch so this is much clearer.

### DIFF
--- a/src/stores/session.ts
+++ b/src/stores/session.ts
@@ -47,8 +47,6 @@ export const useSessionStore = defineStore('session', {
 	getters: {
 		full_name: (state): string => `${state.user?.first_name ?? ''} ${state.user?.last_name ?? ''}`,
 		getUser: (state): User => new User(state.user),
-		student_user_courses: (state) => state.user_courses.filter(c => c.role === 'student'),
-		instructor_user_courses: (state) => state.user_courses.filter(c => c.role === 'instructor'),
 	},
 	actions: {
 		updateExpiry(expiry: number): void {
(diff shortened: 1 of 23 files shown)


pstaabp:
This is ready for review.  Course switching works with no errors now.

src/stores/session.ts
```diff
@@ -45,6 +47,8 @@ export const useSessionStore = defineStore('session', {
 	getters: {
 		full_name: (state): string => `${state.user?.first_name ?? ''} ${state.user?.last_name ?? ''}`,
 		getUser: (state): User => new User(state.user),
+		student_user_courses: (state) => state.user_courses.filter(c => c.role === 'student'),
+		instructor_user_courses: (state) => state.user_courses.filter(c => c.role === 'instructor'),
```
drdrew42, line 51: I don't like this. I only see this being useful for the UserCourses page, and not really anywhere else... so just filter on that one page as we had before.

drdrew42 (changes requested):
I'll put in a PR for this branch -- I think I've got most of these recommendations already in.

drdrew42 (approved):
~~I'll put in a PR for this branch -- I think I've got most of these recommendations already in.~~

I think we've resolved the main issues, but we definitely need to start a list of things to refactor UI-side (axios hooks, that error wrapping function that returns a Promise? etc)

src/stores/set_problems.ts
```diff
@@ -106,9 +106,16 @@ export const useSetProblemStore = defineStore('set_problems', {
...
-			const all_problems = response.data as Array<ParseableProblem>;
-			this.set_problems = all_problems.map((prob) => parseProblem(prob, 'Set') as SetProblem);
+			const response = await api.get(`courses/${course_id}/problems`)
+				.then((response) => {
+					if (response.status !== 200) throw (response.data as Error);
+					return response.data as Array<ParseableProblem>;
+				})
+				.catch((e: Error) => {
+					logger.error(`[fetchSetProblems] ${e.message}`);
+					throw e;
```
drgrice1, line 116: This is really bad.  You are re-throwing the exception that you just caught.  Furthermore when the fetchSetProblems method is called, there is no exception handling.  Any exception that is thrown needs to be handled somewhere higher up in the call sequence, and re-throwing a caught exception is in general a bad idea.

drgrice1 (approved):
This looks okay.  The exception re-throwing needs to be done differently.

src/stores/set_problems.ts
drdrew42: Yeah, this was an accidental inclusion -- it should have been pulled out. There were authentication issues with the stores testing and I inserted this in here to try and capture some info about what was going on. I'll update this PR to remove the catch altogether, see below.

More generally, @pstaabp and I discussed using axios hooks to throw exceptions when responses are not 200 OK. Then we'll need to go through the components and handle exceptions when dealing with store fetches... see #115. Is that along the lines of what you're thinking?

drgrice1: Yeah, that TODO item is what I am thinking!

PR #111, opened 2022-08-10
Remove the firstval function from the List::MoreUtils module

pstaabp:
Remove the firstval function from the List::MoreUtils module.  Also removes the dependencies in the docker files.


pstaabp:
Ready to test now.

drgrice1 (approved):
This pull request looks good.  One less dependency!

drdrew42 (approved):

PR #112, opened 2022-08-18
Refactor delete methods in the database

pstaabp:
This refactors all of the deleteXXX methods in the database layer.  

- Previously `deleteXXX` would return the deleted object.  
- At the db layer, nothing is returned. 
- At the mojolicious layer, a message saying XXX was deleted successfully.

In addition
- all perl tests now delete objects then check that they are successfully deleted. 
- the client-side store tests also delete the object then checks the database for successful deletion. 

Note: The UI never used a deleted object so no UI changes should occur.

### DIFF
--- a/lib/DB/Schema/ResultSet/ProblemPool.pm
+++ b/lib/DB/Schema/ResultSet/ProblemPool.pm
@@ -88,12 +88,12 @@ sub getProblemPool ($self, %args) {
 
 	my $pool = $self->find($search_info, { join => [qw/courses/] });
 
-	unless ($pool) {
-		my $course_name = $course->course_name;
+	DB::Exception::PoolNotInCourse->throw(message => 'The pool with '
+			. ($args{info}{pool_name} ? ' name ' . $args{info}{pool_name} : 'id ' . $args{info}{problem_pool_id})
+			. 'is not in the course '
+			. $course->course_name)
+		unless $pool;
 
-		DB::Exception::PoolNotInCourse->throw(
-			message => "The pool with name \{$args{info}->{pool_name}} is not in the course $course_name");
-	}
 	return $pool if $args{as_result_set};
 	return { $pool->get_columns };
 }
@@ -193,7 +193,7 @@ sub updateProblemPool ($self, %args) {
 
 	my $pool = $self->getProblemPool(info => $args{info}, as_result_set => 1);
 
-	DB::Excpetion::PoolNotInCourse->throw(
+	DB::Exception::PoolNotInCourse->throw(
 		message => 'The problem pool '
 			. (
 				$args{info}->{pool_name}
@@ -312,7 +312,6 @@ This gets a single problem out of a ProblemPool.
 =cut
 
 sub getPoolProblem ($self, %args) {
-
 	my $problem_pool = $self->getProblemPool(info => $args{info}, as_result_set => 1);
 
 	my $pool_problem_info = {};
@@ -328,6 +327,12 @@ sub getPoolProblem ($self, %args) {
 
 	if (scalar(@pool_problems) == 1) {
 		return $args{as_result_set} ? $pool_problems[0] : { $pool_problems[0]->get_inflated_columns };
+	} elsif (scalar(@pool_problems) == 0) {
+		DB::Exception::PoolProblemNotInPool->throw(message => 'The problem with id '
+				. $pool_problem_info->{pool_problem_id}
+				. ' is not in the pool named \''
+				. $problem_pool->pool_name
+				. "'");
 	} else {
 		# Pick a random problem.
 		my $prob = $pool_problems[ rand @pool_problems ];
--- a/lib/DB/Schema/ResultSet/ProblemPool.pm
+++ b/lib/DB/Schema/ResultSet/ProblemPool.pm
@@ -102,6 +102,27 @@ sub getProblemPool ($self, %args) {
 
 Add a problem pool for a given course
 
+=head3 arguments
+
+=over
+
+=item * C<params> a hashref specifying information on the added problem pool
+One must include either the C<course_id> or C<course_name> and the C<pool_name>.
+If there is not enough info to get the course and pool, a C<ParametersNeeded> exception is thrown.
+
+
+=item * C<as_result_set>: boolean
+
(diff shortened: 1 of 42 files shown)


lib/DB/Schema/ResultSet/ProblemPool.pm
```diff
@@ -374,12 +435,8 @@ remove a Problem out of a ProblemPool in a course
...
 
 sub removePoolProblem ($self, %args) {
-	my $prob = $self->getPoolProblem(info => $args{info}, as_result_set => 1);
-	DB::Exception::PoolProblemNotInPool->throw(info => $args{info}) unless defined($prob);
-
-	my $prob2 = $prob->delete;
-	return $prob2 if $args{as_result_set};
-	return { $prob2->get_inflated_columns };
+	$self->getPoolProblem(info => $args{info}, as_result_set => 1)->delete;
+	return;
```
drdrew42, line 439: Not part of this PR, but why is this called `remove` instead of `delete`?

src/stores/problem_sets.ts
```diff
@@ -268,16 +272,21 @@ export const useProblemSetStore = defineStore('problem_sets', {
...
-			// TODO: check for errors
-			const deleted_user_set = parseDBUserSet(response.data as ParseableDBUserSet);
-			const index = this.db_user_sets.findIndex(s => s.user_set_id === deleted_user_set.user_set_id);
-			this.db_user_sets.splice(index, 1);
-			user_set.set(deleted_user_set.toObject());
-			return user_set;
+			if (response.status === 200) {
+				const index = this.db_user_sets.findIndex((set) => set.user_set_id === user_set.user_set_id);
+				if (index < 0) {
+					logger.error('[user store/deleteUserSet]: the user set was not found in the store');
```
drdrew42: Shouldn't we be checking that the user set is present before making the API call? This is an error that 'should never happen', but it doesn't make sense to check afterwards.

Also, I worry about having removed the Exceptions on the DB side for attempting to delete objects that don't exist... that should be an error still, and the response wouldn't be 200 OK.

src/stores/set_problems.ts
```diff
@@ -166,16 +166,18 @@ export const useSetProblemStore = defineStore('set_problems', {
...
+			await api.delete(`courses/${course_id}/sets/${
 				problem.set_id}/problems/${problem.set_problem_id}`);
-			const deleted_problem = new SetProblem(response.data as ParseableSetProblem);
-			// TODO: check for errors
-			const index = this.set_problems.findIndex(prob => prob.set_problem_id === deleted_problem.set_problem_id);
-			this.set_problems.splice(index, 1);
-			return deleted_problem;
+			const index = this.set_problems.findIndex(prob => prob.set_problem_id === problem.set_problem_id);
+			if (index < 0) {
+				logger.error('[stores/set_problems/deleteSetProblem]: the set problem was not found in the store');
```
drdrew42: Check before API

src/stores/set_problems.ts
```diff
@@ -250,26 +252,25 @@ export const useSetProblemStore = defineStore('set_problems', {
...
 			}/users/${user_set.user_id}/problems/${user_problem.user_problem_id}`);
-			const db_deleted_problem = new DBUserProblem(response.data as ParseableDBUserProblem);
+
 			const index = this.db_user_problems
-				.findIndex(user_problem => user_problem.user_problem_id === db_deleted_problem.user_problem_id);
-			if (index >= 0) {
-				this.db_user_problems.splice(index, 1);
+				.findIndex(user_problem => user_problem.user_problem_id === user_problem.user_problem_id);
+			if (index < 0) {
+				logger.error('[stores/set_problems/deleteUserProblem]: the set problem was not found in the store');
```
drdrew42: Check before API

src/stores/users.ts
```diff
@@ -155,15 +155,20 @@ export const useUserStore = defineStore('user', {
...
 			const session_store = useSessionStore();
 			const course_id = session_store.course.course_id;
 			const response = await api.delete(`courses/${course_id}/global-users/${user.user_id ?? 0}`);
 			if (response.status === 200) {
 				const index = this.users.findIndex((u) => u.user_id === user.user_id);
-				// splice is used so vue3 reacts to changes.
-				this.users.splice(index, 1);
-				return new User(response.data as ParseableUser);
+				if (index < 0) {
+					logger.error('[user store/deleteUser]: the user was not found in the store');
```
drdrew42: Check before API

src/stores/users.ts
```diff
@@ -261,16 +266,16 @@ export const useUserStore = defineStore('user', {
...
 			if (response.status === 200) {
 				const index = this.db_course_users.findIndex((u) => u.course_user_id === course_user.course_user_id);
-
-				// splice is used so vue3 reacts to changes.
-				this.db_course_users.splice(index, 1);
-				const deleted_course_user = new DBCourseUser(response.data as ParseableCourseUser);
-				const user = this.users.find(u => u.user_id === deleted_course_user.user_id);
-				return new CourseUser(Object.assign({}, user?.toObject(), deleted_course_user.toObject()));
+				if (index < 0) {
+					logger.error('[user store/deleteCourseUser]: the user was not found in the store');
```
drdrew42: Check before API

drdrew42 (commented):
I'm concerned about lack of Exceptions on delete calls. Roughly speaking, IMO, the approach should be:
1. Fetch the object to be deleted
2. If the object doesn't exist, throw NotFoundException
3. Delete the object

And on the UI side:
1. Check that the object exists in the store
2. If the object doesn't exist, display error message and halt
3. Make API call
4. Handle API response (update store or handle exception message)

lib/DB/Schema/ResultSet/ProblemPool.pm
pstaabp: My thought is that we are removing a problem from a pool, not deleting the problem.  Happy to switch to delete though.

pstaabp:
There are still exceptions thrown.  For example, that you mention above, on `ProblemPool`, the `getProblemPool` will throw an exception if it doesn't exist. This is how all of the deleteXXX methods work as well.

src/stores/problem_sets.ts
pstaabp: Will switch the order of these.  About the exception, see my comment below.

PR #114, opened 2022-08-22
Refactor db relationships

pstaabp:
This changes assessor names in the DBIx::Class relationships for consistency.  For singular relationships, the 's' has been dropped. 

Note: this is based on #106, which has some changes to the CourseSettings database tables.


lib/DB/Schema/Result/Course.pm
```diff
@@ -66,16 +66,16 @@ __PACKAGE__->set_primary_key('course_id');
 
 # set up the many-to-many relationship to users
 __PACKAGE__->has_many(course_users => 'DB::Schema::Result::CourseUser', 'course_id');
-__PACKAGE__->many_to_many(users => 'course_users', 'users');
+__PACKAGE__->many_to_many(users => 'course_users', 'user');
```
drdrew42, line 69: This should remain plural.

lib/DB/Schema/Result/DBPermission.pm
```diff
@@ -63,6 +63,6 @@ __PACKAGE__->add_columns(
 
 __PACKAGE__->set_primary_key('db_perm_id');
 __PACKAGE__->has_many(db_perm_roles => 'DB::Schema::Result::DBPermRole', 'db_perm_id');
-__PACKAGE__->many_to_many(roles => 'db_perm_roles', 'roles');
+__PACKAGE__->many_to_many(roles => 'db_perm_roles', 'role');
```
drdrew42, line 66: This should remain plural.

lib/DB/Schema/Result/Role.pm
```diff
@@ -43,7 +43,9 @@ __PACKAGE__->add_columns(
 __PACKAGE__->set_primary_key('role_id');
 __PACKAGE__->add_unique_constraint([qw/role_name/]);
 
+__PACKAGE__->belongs_to(course_user => 'DB::Schema::Result::CourseUser', 'role_id');
+
 __PACKAGE__->has_many(db_perm_roles => 'DB::Schema::Result::DBPermRole', 'role_id');
-__PACKAGE__->many_to_many(db_perms => 'db_perm_roles', 'db_perms');
+__PACKAGE__->many_to_many(db_permissions => 'db_perm_roles', 'db_permission');
```
drdrew42, line 49: This should remain plural.

lib/DB/Schema/Result/User.pm
```diff
@@ -115,6 +115,6 @@ __PACKAGE__->set_primary_key('user_id');
 __PACKAGE__->add_unique_constraint([qw/username/]);
 
 __PACKAGE__->has_many(course_users => 'DB::Schema::Result::CourseUser', 'user_id');
-__PACKAGE__->many_to_many(courses => 'course_users', 'courses');
+__PACKAGE__->many_to_many(course => 'course_users', 'course');
```
drdrew42, line 118: This should remain plural.

drdrew42 (commented):
Many-to-many relationships are also pluralized.

lib/DB/Schema/Result/Course.pm
pstaabp: I was using the example in https://metacpan.org/pod/DBIx::Class::Relationship#many_to_many

It seems that the 3rd argument is the same as the belongs_to in the bridging table. I don't think this is an accessor name. The example cited above used singular for this relationship (actor and role).

PR #117, opened 2022-08-24
Adapt the unit test action to not use a docker image.

drgrice1:
This just uses the ubuntu 22.04 runner, and installs the webwork3 dependencies directly, and then runs the unit tests.  It actually doesn't take much longer to do this than to use the prebuilt docker image.

This is one way to solve the issue of the dependence on a docker image that does not belong to the openwebwork project.

There is one issue though.  Several of the tests were failing with this new approach.  Looking closely at the tests that were failing, I realized that those tests should be failing.  All of the tests that were failing were boolean values compared to numerical values.  Those tests have been fixed in this pull request.  However, the issue is that for some reason the fixed tests were passing for us when we run the tests on our local machines and they were passing in the previous docker image approach.  We need to figure this out, so that we don't need to wait until the workflow has run to see what needs to be fixed.

The dockerfile used for the previous workflow image has also been updated for new Ubuntu package versions.  Also the dependencies of Devel::Cover::Report::Codecove that are available in the Ubuntu repositories are installed via those packages.  This speeds up the build some.  Ubuntu packages install faster than building from cpan source via cpanm.

Note there was an instance in lib/DB/Schema/Result/ProblemSet.pm of a boolean column, and so that package should load the InflateColumn::Boolean class.  Also, all of the `__PACKAGE__->load_components` calls were made uniform.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -56,7 +56,7 @@ jobs:
       # Run Linter against code base #
       ################################
       - name: Super-Linter
-        uses: github/super-linter@v4.8.1
+        uses: github/super-linter@v4.9.5
 
         env:
           VALIDATE_ALL_CODEBASE: false
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,18 +1,73 @@
+---
 name: Unit Tests and Coverage
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 jobs:
   unit-tests:
-    runs-on: ubuntu-latest
-    # If we are going to use a prebuilt image like this we need a webwork repository on docker hub
-    container: drgrice1/webwork3
+    runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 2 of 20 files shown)


drgrice1:
This pull requests success is self evident.  The unit test check action was successful, and looking back at times that previous unit test actions have taken you will see that it does pretty well.  It took 2 minutes and 51 seconds for the unit test for this pull request.  Looking at previous unit test actions you see that is comparable.  I have run this numerous times on my fork and this is consistent.

drgrice1:
The docker image built with the `webwork3.dockerfile` in this pull request will now also have the stricter boolean type checking.  So the unit tests and the docker image are consistent, and the unit tests will fail in both without the changes to the tests in this pull request.

It would be nice to force the tests to use a specific JSON backend module, regardless of what is installed on the system.  The environment variables PERL_JSON_BACKEND and MOJO_NO_JSON_XS should work for this, but they don't for some reason.

drgrice1:
I added the Cpanel::JSON::XS module to the workflow and to the webwork2.dockerfile.  So now the tests will run the way they do now on our local setups.

I will work to convert the tests to Test2::V0 and Test2::MojoX to make the tests more strict for boolean types.

Issue #110, opened 2022-08-07
WeBWorK 3 backend to-do list


drdrew42:
- [ ] refactor relationships for plural/singular relation names
- [ ] update all instances of validation to use new `Validation.pm`
- [ ] use `->insert` wherever `->new(...)` is used
- [ ] add `fetch_with_defaults` as an alias for `discard_changes` and employ everywhere `add_to_$rel` is used
- [ ] eliminate dependency on `List::MoreUtils` (~~firstval~~)
- [ ] update all `$self` to use more appropriate naming convention

drdrew42:
@pstaabp @drgrice1 feel free to add to this checklist 🙏

pstaabp:
I just edited the list above.

drdrew42:
`discard_changes` is no longer needed, now that we've added `retrieve_on_insert => 1` to all columns with default values. Please keep this in mind for any future DB changes that involve default values.

PR #119, opened 2022-08-28
Switch to using the Test2::Suite and Test2::MojoX modules for unit tests.

drgrice1:
Note that is important for the unit tests to have the `got` argument first and the `expected` argument second in an `is` test.  Test2::Tools::Compare converts the second `expected` argument into a Test2::Tools object which is important for strict checks.
    
Also remove the incomplete t/db/test_course_user.pl file.

Note that this depends on #117.


drgrice1:
Note that this requires version 0.000139 or newer of Test2::Suite since it uses check_isa in the tests.  The version on Ubuntu 20.04 is 0.000129, so an install from cpan is needed there.  The version on Ubuntu 22.04 is 0.000144, and so is new enough.  I need to upgrade.

t/db/004_course_users.t
```diff
@@ -161,90 +165,78 @@ my $quimby_params = clone($course_user_params);
...
-			}
-		}
-	);
-}
-'DB::Exception::InvalidParameter', 'addCourseUser: an parameter with invalid value';
+is(
+	dies {
+		$user_rs->addCourseUser(
+			info   => { course_name => 'Topology', username           => 'quimby' },
+			params => { role        => 'student',  course_user_params => { useMathQuill => 0 } }
```
pstaabp, line 222: useMathQuill is supposed to be a boolean.  Switch to false?

t/mojolicious/002_courses.t
```diff
@@ -124,8 +124,8 @@ $t->get_ok('/webwork3/api/courses/4/settings')->status_is(200)->content_type_is(
 $t->post_ok('/webwork3/api/courses' => json => $new_course)->status_is(403)->json_is('/has_permission' => 0);
 
 $t->put_ok('/webwork3/api/courses/4' => json => { course_name => 'XXX' })->status_is(403)
-	->json_is('/has_permission' => false);
+	->json_is('/has_permission' => 0);
```
pstaabp, line 127: This seems like this should be a boolean and can then this with a switch on line 95 in `WeBWorK3::Controller::Permission` to `has_permission => false`.

pstaabp: This would then affect almost all of the other mojolicious tests as well though.

t/db/005_hwsets.t
```diff
@@ -275,57 +255,62 @@ my $new_set7 = {
...
-	);
-}
-'DB::Exception::InvalidParameter', 'addProblemSet: adding an non-valid boolean parameter in set_dates';
+is(
+	dies {
+		$problem_set_rs->addProblemSet(
+			params => {
+				course_name => 'Precalculus',
+				set_name    => 'HW #11',
+				set_dates   => { open => 100, due => 140, answer => 200, enable_reduced_scoring => 0 },
```
pstaabp, line 305: `enable_reduced_scoring => false`

pstaabp:
Overall, this looks great.  Good cleanup and improved functionality.  Just added a couple of minor items in the code.

t/db/004_course_users.t
drgrice1: This is supposed to be 0.  That is the point of this test.  It is checking that an invalid field is passed in the parameters.  If this is switched to `false` an exception will not be thrown because that is valid.

t/mojolicious/002_courses.t
drgrice1: I am not going to change this in this pull request.  This was changed to 0 because currently the method returns a numeric (not boolean) value.  That is a TODO for later.

t/db/005_hwsets.t
drgrice1: Again, this needs to be 0.  That is the point of this test.

drgrice1: Although, it seems that `set_params => { hide_hint => 0 }` should be `false` instead of `0` in this test so that the `enable_reduced_scoring` value is appropriately tested (the previous test already checks the `set_params` values.

drdrew42:
It sounds like we will need a separate PR that standardizes the use of booleans (and corresponding cleanup to the DB to ensure that our boolean fields are properly typed). I'll add this to #110

drgrice1:
I can make the change to permissions in this pull request if you want me to.  However, the objective of this pull request was to make the switch to Test2::Suite and Test2::MojoX, and not to change the current behavior.  I had observed that the `has_permission` method was returning a numeric value, and  not a boolean.  I was going to point this out, and suggest that it be changed.  @pstaabp clearly noticed this as well.   So the more strict tests have already revealed something needed!

pstaabp:
I'm fine merging in and then doing some cleanup, especially with booleans. That's probably the better way to go.

PR #120, opened 2022-08-28
Consolidate the login code to the session store

pstaabp:
When logging in if the user is authenticated, there are still a number of things that occur
1. storing the user in the session
2. fetching user courses
3. fetching roles & permissions

Although this only needs to be done once, the store tests should be logging in as a non-admin for better testing and these steps need to be done for each test.  Therefore, all of these steps are pulled into a single login action in the session store.

### DIFF
--- a/src/common/models/users.ts
+++ b/src/common/models/users.ts
@@ -6,15 +6,19 @@ import { isNonNegInt, isValidUsername, isValidEmail }
 import { Dictionary, Model } from 'src/common/models';
 import { UserRole } from 'src/stores/permissions';
 
-export interface ParseableUser {
-	user_id?: number;
-	username?: string;
+// This defined the required fields of a user in the Session
+
+export interface SessionUser {
+	user_id: number;
+	username: string;
 	email?: string;
 	first_name?: string;
 	last_name?: string;
-	is_admin?: boolean;
+	is_admin: boolean;
 	student_id?: string;
 }
+
+export type ParseableUser = Partial<SessionUser>;
 
 /**
  * @class User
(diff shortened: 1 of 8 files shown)


src/common/models/users.ts
```diff
@@ -34,15 +38,15 @@ export class User extends Model {
 	get all_field_names(): string[] { return User.ALL_FIELDS; }
 	get param_fields(): string[] { return []; }
 
-	constructor(params: ParseableUser = {}) {
+	constructor(params: ParseableUser = { user_id: 0, username: '', is_admin: false }) {
```
drdrew42, line 47: username: `logged_out`?

Perhaps export this default as a constant for use in session.ts?

src/stores/session.ts
```diff
@@ -25,18 +25,20 @@ interface CourseInfo {
 export interface SessionState {
 	logged_in: boolean;
 	expiry: number;
-	user: ParseableUser;
+	user: SessionUser;
 	course: CourseInfo;
 	user_courses: ParseableUserCourse[];
 }
 
+const logged_out_user = { username: 'logged_out', user_id: 0, is_admin: false };
```
drdrew42: import default user from `model/users.ts`

tests/stores/session.spec.ts
```diff
@@ -39,14 +38,10 @@ describe('Session Store', () => {
 		is_admin: false,
 	};
 
-	const logged_out: ParseableUser = {
+	const logged_out: SessionUser = {
```
drdrew42: import from `model/users.ts`

drdrew42 (commented):
Create and export a constant for the default user -- to be used in testing, login failure, and logout.

src/common/models/users.ts
pstaabp: That's a good idea

PR #122, opened 2022-11-16
Unit test rework

drgrice1:
I am just quickly putting this in for you to see.  This commentary needs work.  See `git log` for more details.


pstaabp:
Everything runs well (but slow) with this.  It seems like with every test run, the db is started up and loaded with data.  Nice that each time there is a fresh data available and you know the state of the database, I guess this method you don't need to check that the database is restored to its previous state.   Is there anyway of keeping the database up for multiple tests? I'm guessing that things would run a lot faster (if that matters). 

I think this could go in and rebase other branches off this--that'll take some work.  Clearly all of the tests have been changed, but only some minor cosmetic changes to any of the core code.  I tested the ww3 interface and seems to work as before.

pstaabp (approved):

drgrice1:
Yes, it will run slower than before.  It is doing a lot more testing.

Yes, every test starts with a fresh new database.  There is no need to restore the database to some previous state because of this.  There is a way to keep the database up for multiple tests, but that would revert back to before.  That is not what is desired.

The point of unit tests is not to be fast.  It is to check that things are working correctly.

drgrice1:
You have to realize that the old approach of trying to put the database back into its original state is unreliable and unmaintainable.  That is the point of each test using a fresh new database built from scratch.  That makes the tests take longer, but it is worth it.

pstaabp:
This makes sense after thinking about it more.  And the time slowdown shouldn't been too bad--one can run the individual tests fairly quickly on a single database and be patient when running all of them.

It's also fantastic that all 3 databases can be tested easily.  Should catch some of those errors that popped up before.

duffee:
You might be interested in using fixtures for mocking database tests.  In a dbic environment, we use [Test::DBIx::Class](https://metacpan.org/pod/Test::DBIx::Class) and [DBIx::Class::Fixtures](https://metacpan.org/pod/DBIx::Class::Fixtures) something like this
```perl
use Test::DBIx::Class {
    schema_class => 'Something::Something::DB'
};

sub connect {
    ...
    DBIx::Class::Fixtures->new( {options => ...} )->populate( {...} );
}
```

I haven't been able to find an example of a non-dbic setup, but maybe something like [Test::mysqld](https://metacpan.org/pod/Test::mysqld) or even [DBIx::Class::Migration](https://metacpan.org/pod/DBIx::Class::Migration) might give you the database tests you're looking for without the overhead of creating the full-blown database everytime.  Just a thought.

